PR #2, opened 2020-10-06
Implemented bubbles list widget

treilik:
Implemented a list widget upon the viewport-widget to view and select list items.

Its not yet able to handle strings containing words longer than the viewport width and highlighting is also not bug free.


meowgorithm:
This is cool! Let us know if you need any help with it.

treilik:
So, after a long time, I am now happy with the code, the history and the tests for a review.
Even if there are still things i want to improve:
- [ ] add `HardWrap` after it is [available](https://github.com/muesli/reflow/pull/3) to fix to long lines.
- [ ] check in the test-file, not only if error, but also for the Type of the error, if there should be an error.
- [ ] add more (golden sample) tests and also cover more edge cases within the existing tests.
- [ ] change example item struct to have a unique id to properly set a equals function, that don't rely on string comparison.
If you have any thoughts or change request please do let me know.
Cheers!

list/list_test.go
```diff
@@ -143,18 +115,17 @@ func TestMultiLineBreaks(t *testing.T) {
 	m.SuffixGen = NewSuffixer()
 	m.Screen = ScreenInfo{Height: 50, Width: 80}
 	m.AddItems(MakeStringerList([]string{"\n\n\n\n\n\n\n\n\n\n\n\n\n\n\n\n\n\n\n\n\n\n\n\n\n\n\n\n\n\n\n\n\n\n\n\n\n\n\n\n\n\n\n\n\n\n\n\n\n\n\n\n\n\n\n\n\n\n\n\n"}))
```
quenbyako, line 122: Wait, WHAT??? o_O

treilik: Well, one has to test all contingencies :) 
For example a item which has more lines than the screen has... dont you think?

PR #24, opened 2020-10-28
Return the value of paginator.TotalPages in SetTotalPages

penguwin:
...when the given number of items is below 1.


meowgorithm:
Pengu: fixing meowgorithm logic mistakes since 2020™

PR #26, opened 2020-10-30
Add vertical layout

muesli:
A simple vertical layout that can take an arbitrary amount of children, which you can switch focus between by pressing tab.


maaslalani:
Cleaning up some old PRs, feel free to reopen if we still want to add this.

PR #30, opened 2020-11-04
Feat: added progress bar

ololosha228:
Awesome. Colorful.

Well, docs are still work in progress, but it's already looks pretty cool, i think.


muesli:
Oops, looks like an empty file was committed?

ololosha228:
@muesli  oh crap! fixed

Looks like i messed up a bit with too old version

progress/progress.go
```diff
@@ -0,0 +1,91 @@
...
+
+type Model struct {
+	// Left side color of progress bar. By default, it's #00dbde
+	StartColor colorful.Color
+
+	// Left side color of progress bar. By default, it's #fc00ff
+	EndColor colorful.Color
+
+	// Length of progress bar in symbols.
+	Length int
```
meowgorithm, line 22: I'd change this to `Width` for consistency with other packages in this repo.

ololosha228: Yeah, i agree

ololosha228:
okay okay @muesli @meowgorithm i fixed it by my hands)))

progress/progress.go
```diff
@@ -43,7 +43,7 @@ func NewModel(size int) *Model {
 	return &Model{
 		StartColor:     startColor,
 		EndColor:       endColor,
-		Length:         size,
+		Width:         size,
```
muesli, line 46: Please make sure your code is formatted with `gofmt`. I can recommend enabling that as a save-hook in your editor/IDE.

ololosha228: I edited via github ide, sorry, i didn't note that

muesli: Don't worry, it happens :smiley:

ololosha228:
@muesli @meowgorithm is this PR looks good?

meowgorithm:
Hey! Sorry to let this one sit. This is a great start, but we'd like to flesh out this component a bit more before adding it to the library. For example, adding an option for some numeric progress (i.e. `78%`)

I'll make some updates to your branch.

meowgorithm:
Hey, just a note that I've started looking at this. I'll push to your branch in a little bit.

meowgorithm:
Okay, I think that's everything. Unless you have any feedback on this, I'll merge it in.

And thanks for getting this going. I really like the details you added with respect to the gradient.

ololosha228:
@meowgorithm yeeeeaaah, i have a few:

Main question is why multiple deps with color stuff? I mean, emmmm `github.com/lucasb-eyer/go-colorful` does pretty same thing as `github.com/muesli/termenv`))))

Also, is hex colors as color input a good idea? May be we can unify any color stuff in bubbletea using std image/color.Color, and create simplest wrapper, which can make ansi escape codes from rgba func? 

Other questions i'll add below

progress/progress.go
```diff
@@ -0,0 +1,192 @@
...
+	// "Empty" sections of progress bar
+	Empty      rune
+	EmptyColor string
+
+	// Settings for rendering the numeric percentage
+	ShowPercentage  bool
+	PercentFormat   string // a fmt string for a float
+	PercentageStyle *termenv.Style
+
+	useRamp    bool
```
ololosha228, line 96: maybe we should make it public?

progress/progress.go
```diff
@@ -0,0 +1,192 @@
...
+
+// WithDefaultScaledRamp sets a gradient with default colors, and scales the
+// gradient to fit the filled portion of the ramp.
+func WithDefaultScaledRamp() Option {
+	return WithScaledRamp("#00dbde", "#fc00ff")
+}
+
+// WithScaledRamp scales the gradient to fit the width of the filled portion of
+// the progress bar.
+func WithScaledRamp(colorA, colorB string) Option {
```
ololosha228, line 44: That's called "gradient" actually, "ramp" term can be ambiguous

progress/progress.go
```diff
@@ -0,0 +1,192 @@
...
+	var (
+		tw = m.Width - textWidth        // total width
+		fw = int(float64(tw) * percent) // filled width
+		p  float64
+	)
+
+	if m.useRamp {
+		// Gradient fill
+		for i := 0; i < fw; i++ {
+			if m.scaleRamp {
```
ololosha228, line 153: Looks  like `m.scaleRamp` literally does nothing :smile:

meowgorithm: The lines are similar looking but not the same:

```go
tw = m.Width - textWidth        // total width
fw = int(float64(tw) * percent) // filled width

p = float64(i) / float64(fw) // first case
p = float64(i) / float64(tw) // second case
```

progress/progress.go
meowgorithm: The terms are fairly interchangeable; "gradient" is more common with web designers while "ramp" is more common in motion graphics. You have a point, though: most users in this case will probably be more familiar with the term "gradient."

progress/progress.go
meowgorithm: I'm hesitant to make it public because it also requires generating the two `colorful` colors that go with it. In that case we could use a setter function, but I figure at that point one may as well just recreate the struct with `NewModel()`.

meowgorithm:
> Main question is why multiple deps with color stuff? I mean, emmmm `github.com/lucasb-eyer/go-colorful` does pretty same thing as `github.com/muesli/termenv`))))

`go-colorful` is for generating colors, whereas `termenv` is for applying ANSI styling in the terminal. The very important thing that `termenv` does is gracefully degrade colors for terminals (and situations) where 24-bit color isn't supported. For example, the macOS default terminal only supports 256 colors (iirc) so termenv will gracefully convert the colors to the closest value supported or, if output is not a terminal (like if the output is going through a pipe), `termenv` will disable styling altogether. It's a really great package; it does some other helpful things too. @muesli, one of the authors of this project, maintains the package.

> Also, is hex colors as color input a good idea? May be we can unify any color stuff in bubbletea using std image/color.Color, and create simplest wrapper, which can make ansi escape codes from rgba func?

I find hex colors quite easy to work with but, yeah, there's definitely some potential for using `color.Color` as a standard. We'd still probably use `termenv` regardless because of the way it gracefully handles colors, though.

muesli:
> Also, is hex colors as color input a good idea? May be we can unify any color stuff in bubbletea using std image/color.Color, and create simplest wrapper, which can make ansi escape codes from rgba func?

Yeah, that's probably a good idea. `colorful`'s colors are already compatible with that interface, if I remember correctly.

ololosha228:
@muesli off-topic, but what do you think about some unified stylesheets or, if you want, 'themes' for bubbles?

I mean, we have absolutely amazing flutter framework, and i think bubbletea and bubbles has pretty same concept, that 'it's all about ~~widgets~~ bubbles', so, what if we can copy some concepts like stylesheets, ~~MaterialApp~~ no we don't want copy this shit, by another things are sooo useful in cli apps too.

I currently working on closed lib, which supports css themes for ansi colors, maybe i need make it opensource

muesli:
@ololosha228 Yeah, we'll definitely need some form of a color palette & theme. `gamut` already provides a few types we can experiment with.

ololosha228:
Niiice, thanks!

Issue #25, opened 2020-10-30
Select, multi-select and multiline answer components


chriswalz:
Also a multiline answer.  I would definitely see myself using them on https://github.com/chriswalz/bit. 

Right now I'm using https://github.com/AlecAivazis/survey

muesli:
Agreed, these are desperately needed as a standard bubbles and we're working on it!

Just as a heads up: in the meantime you can already find a custom select example in bubble tea: https://github.com/charmbracelet/bubbletea/blob/master/examples/views/main.go

erikgeiser:
I already implemented a select-component (not mutli-select though) based on `bubbletea` over here: https://github.com/erikgeiser/promptkit

It has the following features:
* Pagination
* Filtering
* Custom styling (with color/style support via https://github.com/muesli/termenv)
* Custom choice types
* Proper wrapping for small terminals via https://github.com/muesli/reflow

Currently it is not ready because:
* It depends on my PR in `reflow`: https://github.com/muesli/reflow/pull/30 (currently I use a local redirect in the `go.mod`)
* I just finished it and didn't test it exhaustively
* `SliceChoices` only works with slice of pointers for some reason

If you are interested I can submit this component as a PR to https://github.com/charmbracelet/bubbles and continue to develop it over here.

Here are some examples:
[![asciicast](https://asciinema.org/a/4ZK5HZ2uJm8NtC0rs8rnqxUwS.svg)](https://asciinema.org/a/4ZK5HZ2uJm8NtC0rs8rnqxUwS)
[![asciicast](https://asciinema.org/a/HZ7BOXUgOsFx7u5p1eId9I33b.svg)](https://asciinema.org/a/HZ7BOXUgOsFx7u5p1eId9I33b)

erikgeiser:
@muesli what do you think? Do you want a PR or are you working on your own implementation?

muesli:
Looking good! We love PRs :heart: and I think it makes sense to establish a few default components.

maaslalani:
@chriswalz I know this is quite an old issue but we've just introduced a new multi-line input in case you still wanted one 🙂 

<img src="https://stuff.charm.sh/bubbles-examples/textarea.gif" width="600" />

It's available on `master` https://github.com/charmbracelet/bubbles and there are some examples on how to use it here https://github.com/charmbracelet/bubbletea/pull/357

maaslalani:
[...]

PR #33, opened 2020-11-28
Add selection prompt

erikgeiser:
This PR adds a selection prompt with the following features:
* Optional Pagination
* Optional Filtering
* Customizable appearance (with color/style support via https://github.com/muesli/termenv)
* Custom choice types
* Proper wrapping for small terminals via https://github.com/muesli/reflow

Examples:
[asciicast](https://asciinema.org/a/4ZK5HZ2uJm8NtC0rs8rnqxUwS)
[asciicast](https://asciinema.org/a/HZ7BOXUgOsFx7u5p1eId9I33b)

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -6,6 +6,7 @@ require (
 	github.com/atotto/clipboard v0.1.2
 	github.com/charmbracelet/bubbletea v0.12.2
 	github.com/mattn/go-runewidth v0.0.9
+	github.com/muesli/reflow v0.2.1-0.20201126184510-3bcb929042f2
 	github.com/muesli/termenv v0.7.4
 	golang.org/x/crypto v0.0.0-20201016220609-9e8e0b390897 // indirect
 	golang.org/x/sys v0.0.0-20201020230747-6e5568b54d1a // indirect
--- a/go.sum
+++ b/go.sum
@@ -12,6 +12,8 @@ github.com/mattn/go-isatty v0.0.12 h1:wuysRhFDzyxgEmMf5xjvJ2M9dZoWAXNNr5LSBS7uHX
 github.com/mattn/go-isatty v0.0.12/go.mod h1:cbi8OIDigv2wuxKPP5vlRcQ1OAZbq2CE4Kysco4FUpU=
 github.com/mattn/go-runewidth v0.0.9 h1:Lm995f3rfxdpd6TSmuVCHVb/QhupuXlYr8sCI/QdE+0=
 github.com/mattn/go-runewidth v0.0.9/go.mod h1:H031xJmbD/WCDINGzjvQ9THkh0rPKHF+m2gUSrubnMI=
+github.com/muesli/reflow v0.2.1-0.20201126184510-3bcb929042f2 h1:+cpkcmASpeBn4qXz2tU+x+njyKe91NoHXqrJdhoDnZo=
+github.com/muesli/reflow v0.2.1-0.20201126184510-3bcb929042f2/go.mod h1:qT22vjVmM9MIUeLgsVYe/Ye7eZlbv9dZjL3dVhUqLX8=
 github.com/muesli/termenv v0.7.2 h1:r1raklL3uKE7rOvWgSenmEm2px+dnc33OTisZ8YR1fw=
 github.com/muesli/termenv v0.7.2/go.mod h1:ct2L5N2lmix82RaY3bMWwVu/jUFc9Ule0KGDCiKYPh8=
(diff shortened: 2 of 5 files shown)


meowgorithm:
Thanks for the PR! At first glance this looks pretty good. Could you add an example application in a subdirectory (perhaps called `example`)?

meowgorithm:
Actually, never mind; found your PromptKit project 😄

erikgeiser:
My bad, I forgot to add the asciinema recordings I made for #25.

There's one concern I have before merging. This prompt supports custom key mappings, and a default key map is used when instantiating it via `NewModel()`. However, if a user manually instantiates it (`m := selection.Model{}`) it will not react to user input at all when no key map is set and the program has to be killed. Is that a user error or should I add some logic to detect whether the bare minimum set of keys is defined in the key map (maybe up, down, select). If not it would throw an error on startup.

Now, when writing this down, I think I should definitely do this.

erikgeiser:
Another consideration: Should I change the `NewModel` signature to take the choices as an argument? Currently, I chose the `NewModel` signature without arguments to be in line with the other models in this package. However, given that the choices are obviously mandatory it would make sense to take it as an argument such that `NewModel` returns a working prompt.

meowgorithm:
Yes, that makes sense; so maybe something like `func NewModel(choices ...[]string)` to keep it light. Another option might be to use functional arguments to allow for different ways of constructing the choices:

```go
m := NewModel(WithChoices("a", b"))
```

This would also allow other options such as:

```go
m := NewModel(
    WithChoices(c),
    WithKeybindings(k),
)
```

erikgeiser:
I would have used `NewModel(choices []*Choices)`. This would then be used like this: `NewModel(SliceChoices(mySlice))`. This would cover all use cases and not just strings. In my opinion this is important because often the choices are arbitrary objects like in the custom example (see also second video above):

```go
choices := []Article{
    {ID: "123", Name: "Article A"},
    {ID: "234", Name: "Article B"},
    {ID: "345", Name: "Article C"},
    {ID: "456", Name: "Article D"},
    {ID: "567", Name: "Article E"},
}

prompt := selection.NewModel(selection.SliceChoices(choices))
prompt.Label = "Choose an article!"
prompt.PageSize = 3
prompt.Template = customTemplate
prompt.Filter = func(filter string, choice *selection.Choice) bool {
    article, _ := choice.Value.(Article)

    return strings.Contains(article.ID, filter)
}

```

I think that your second option would be too much given that the user can just set the optional configuration directly on the model. This also feels more in line with the other models. The choices would be taken as an argument only because they are mandatory.

erikgeiser:
Is there anything I can do or change to move forward with this?

meowgorithm:
@erikgeiser Sorry for the delay on this one; we've been really busy over here. It may be a little while until we can give this a proper review, but we will get to it.

erikgeiser:
Do you still plan to do something about this PR?

meowgorithm:
Hey Erik. We looked at this again, and while the component is a bit too specific for this repository, we really like it. Would you be open to creating a repository for it so we can link to it from the README as an additional resource?

erikgeiser:
Okay, I can work with that decision. I'll see if I can find the time to make a small prompt library around the selection prompt.

meowgorithm:
Sounds good. Please do let us know when it's ready.

I enjoyed reading your code, by the way. They notion of a key map you came up with was particularly impressive.

erikgeiser:
Thank you.

Ultimately, I think it is a good decision not to merge this into `bubbles`. Looking back I developed it to be directly used as a prompt by users while most other `bubbles` components are more like widgets that are intended to be used in larger models.

I took this opportunity to create a prompt library on top of `bubbles` containing this selection prompt:

https://github.com/erikgeiser/promptkit

meowgorithm:
It's great to see another release of promptkit and nice that the library works both standalone and as something that can be integrated into a larger Bubble Tea project.

I've added promptkit to the [README](https://github.com/charmbracelet/bubbles#additional-bubbles).

erikgeiser:
Thank you, currently the library is fairly limited due to https://github.com/charmbracelet/bubbletea/issues/24 as you can only run one prompt before key strokes are lost, but I'll see what I can due about that issue.

meowgorithm:
Ah, good point. It's an interesting issue and a tricky one to solve from what I understand. There are some notes in the thread around it: you're more than welcome to take a look.

muesli:
Gonna close this one, as we resolved the situation. Thank you for all your amazing work, @erikgeiser!

PR #38, opened 2021-02-26
Persist placeholder suffix if input matches placeholder in textinput

fredliang44:
Here is the sample

Before:

https://user-images.githubusercontent.com/22412638/109347774-a3a3dd80-78ae-11eb-8c60-47ab28b73a8e.mp4

After

https://user-images.githubusercontent.com/22412638/109347803-ab638200-78ae-11eb-83bd-08928e7f1e45.mp4

### DIFF
--- a/textinput/textinput.go
+++ b/textinput/textinput.go
@@ -560,11 +560,6 @@ func (m Model) Update(msg tea.Msg) (Model, tea.Cmd) {
 
 // View renders the textinput in its current state.
 func (m Model) View() string {
-	// Placeholder text
-	if len(m.value) == 0 && m.Placeholder != "" {
-		return m.placeholderView()
-	}
-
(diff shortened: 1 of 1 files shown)


maaslalani:
Hey! This is awesome UX but I _think_ this might not be the correct way to implement this feature. I think placeholder's should definitely be cleared once the textinput is not empty any more **but** I would really like to see the placeholder possibly be provided to the user as a suggestion which could get implemented in #140. This is because most placeholder text would not be an actual value the user should input.

I.e. once the user has implemented some text an autocompletion suggestion will suggest the placeholder text (if it is a valid input) which is much more flexible.

I really appreciate this and it is a great addition and hopefully we will have a way to emulate this eventually 🙂 but for now I will close this so that we can focus on the autocomplete / suggestions feature!

PR #43, opened 2021-03-12
Drop naked returns in textinput model

muesli:
This actually uncovered a few ineffective assignments, particularly in deleteWordRight.


meowgorithm (approved):
Looks good and makes sense to me. Thanks, Muesli.

PR #67, opened 2021-09-04
feat: simple timer component

### LINKED ISSUE #22: Timer component
👋🏽 hi, awesome work on the whole bubbletea framework!

### What
I was wondering what would be the `tea`  semantic way to incorporate a timer into a tea application. I want the app to display the time that is left and for the it to exit gracefully once the timer is done. 

I looked under the hood of the Spinner, given that it ticks at at regular interval, as a base for a ticker component. I spotted the [`tea.Tick`](https://github.com/charmbracelet/bubbletea/blob/9e9e05774cb6583783fad87ea73f8e0b696bd75c/commands.go#L48) function which looks promising.

caarlos0:
closes #22 

A simple timer component:

- ticks every second
- on timeout, executes the given `tea.Cmd`

Simplest possible usage example:

```go
package main

import (
	"fmt"
	"os"
	"time"

	"github.com/charmbracelet/bubbles/timer"
)

func main() {
	m := timer.Model{
		Timeout: 10 * time.Second,
		OnTimeout: func() tea.Cmd {
			return tea.Quit
		},
	}

	if err := tea.NewProgram(m).Start(); err != nil {
		fmt.Println("Oh no, it didn't work:", err)
		os.Exit(1)
	}
}
```

let me know what you think

### DIFF
--- a/timer/timer.go
+++ b/timer/timer.go
@@ -46,7 +46,7 @@ func (m Model) View() string {
 }
 
 func (m Model) tick() tea.Cmd {
-	return tea.Every(time.Second, func(_ time.Time) tea.Msg {
+	return tea.Tick(time.Second, func(_ time.Time) tea.Msg {
 		return tickMsg{}
 	})
 }
--- a/timer/timer.go
+++ b/timer/timer.go
@@ -1,4 +1,4 @@
-// Package timer provides a simple timeout component that ticks every second.
+// Package timer provides a simple timeout component.
 package timer
 
 import (
--- a/timer/timer.go
+++ b/timer/timer.go
@@ -1,0 +1,52 @@
+// Package timer provides a simple timeout component that ticks every second.
+package timer
+
+import (
+	"time"
+
+	tea "github.com/charmbracelet/bubbletea"
+)
+
+type tickMsg struct{}
+
+// Model of the timer component.
+type Model struct {
+	// How long until the timer expires.
+	Timeout time.Duration
+
+	// What to do when the timer expires.
+	OnTimeout func() tea.Cmd
+}
+
+// Init starts the timer.
+func (m Model) Init() tea.Cmd {
+	return m.tick()
+}
+
+// Update handles the timer tick.
+func (m Model) Update(msg tea.Msg) (tea.Model, tea.Cmd) {
+	switch msg.(type) {
+	case tickMsg:
+		m.Timeout -= time.Second
+		if m.Timeout <= 0 {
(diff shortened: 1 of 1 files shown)


caarlos0:
one thing that comes to mind now is: should we have an "increasing timer" as well? if so, maybe this component should be called `timeout` or something like that

meowgorithm:
This is cool!

* I'd expose `tickMsg` so users can respond to ticks if they want
* It should probably also send a `TimeoutMsg` on timeout so that users have the option of listening for timeouts without running a command
* It would be cool if `View` returned the time remaining (otherwise it won't ever change)

It would also be useful to allow for a custom decrement value, so you could do something like:

```go
m := Timer{
	Timeout: time.Millisecond * 250,
	TickEvery: time.Millisecond, // or maybe it's called Decrement?
}
```

…and then maybe if the decrement value is `0` it defaults to `time.Second`?

Otherwise, I think `Timer` is a good name for this since it's similar to the functionality in `time.Timer`. I'd probably call an increasing one `Stopwatch`, which I _think_ makes sense to include.

caarlos0:
agreed, made the changes...

view actually returns the time remaining because on update we do `timeout =- interval`

an example usage now looks like:

```go
package main

import (
	"fmt"
	"os"
	"time"

	"github.com/charmbracelet/bubbles/timer"
)

type model struct {
	timer timer.Model
}

	return m.timer.Init()
}

	return "Will exit in " + m.timer.View()
}

	switch msg.(type) {
	case timer.TimeoutMsg:
		return m, tea.Quit
	}
	var cmd tea.Cmd
	m.timer, cmd = m.timer.Update(msg)
	return m, cmd
}

func main() {
	m := model{
		timer: timer.New(3*time.Second),
	}

	if err := tea.NewProgram(m).Start(); err != nil {
		fmt.Println("Oh no, it didn't work:", err)
		os.Exit(1)
	}
}
```

meowgorithm:
LGTM!

timer/timer.go
```diff
@@ -0,0 +1,68 @@
...
+// TimeoutMsg is a message that is sent once when the timer times out.
+type TimeoutMsg struct{}
+
+// Model of the timer component.
+type Model struct {
+	// How long until the timer expires.
+	Timeout time.Duration
+
+	// How long to wait before every tick. Defaults to 1 second.
+	TickEvery time.Duration
```
meowgorithm, line 22: Actually—this should probably be `Interval` to match the other language, which is frankly better.

PR #77, opened 2021-11-09
expose list.FilterMachesMsg

jon4hz:


meowgorithm:
Thanks for this @jon4hz!

For those tuned in, some background here: we're generally erring on exposing internal messages now to make it easier to avoid avoid to route all messages through various Bubbles. In other words, catch-all updates like this:

```go
    // Pass every single message through the list's update function
    var cmd tea.Cmd
    m.list = m.list.Update(msg)
}
```

Can now also be avoided if desired:

```go
    switch msg.(type) {
        case list.FilterMsg:
            var cmd tea.Cmd
            m.list = m.list.Update(msg)
            return m, cmd
    }
}
```

The former example is still totally fine and totally valid, but by exposing internal messages we're now able to be more flexibleand specific in our update functions if need be.

Issue #56, opened 2021-07-09
Truncate at W\idth


meowgorithm:

mrusme:
Is this the reason why `SetcContent` with a very large text will make the viewport expand in height? I was under the assumption that I might have done something wrong, so I'm still looking into why the viewport won't obey the height I set. I thought that a large content will make the viewport scroll vertically.

meowgorithm:
@mrusme iirc, yes. you'll want to wrap your text with something like:

```go
str := lipgloss.NewStyle().Width(width).Render(content)
model.SetContent(str)
```

captv89:
@meowgorithm When the length of a line exceeds the width of the terminal, the line is soft-wrapped as per your example above. However, this is affects the scrolling. Scrolling is still assuming that each content line = line in the terminal, and stops after the YOffset has been incremented to the int value, which is the difference between viewport lines and total content lines. This affects the scroll percentage as well.

Subilan:
Not sure if it's relevant here but it would be nice if something like
```go
str := lipgloss.NewStyle().Width(width).Render(content)
model.SetContent(str)
```
can do natural CJK character wrapping as well. For now, as an example, it seems that the underlying logic considers a whole Chinese sentence as a single word and makes some sort of non-breaking-word decision.

Figure 1: the sentence is moved to another line as a whole
<img width="652" height="58" alt="Image" src="https://github.com/user-attachments/assets/170d3def-4637-4715-b7cf-5ec8be2caf59" />

Figure 2: Claude Code behaviour which is favorable in this case
<img width="646" height="329" alt="Image" src="https://github.com/user-attachments/assets/767fb7a7-5d51-4110-9862-399a7c24d475" />

PR #87, opened 2021-12-25
allow viewport to wrap lines

IllusionMan1212:
This PR addresses issue #56 (partially? not sure about the truncation part).

Lines should wrap according to the width of the viewport. resizes can be handled by the user easily by updating the viewport width when `tea.WindowSizeMsg` is fired.

### DIFF
--- a/viewport/viewport.go
+++ b/viewport/viewport.go
@@ -7,6 +7,27 @@ import (
 	"github.com/charmbracelet/bubbles/key"
 	tea "github.com/charmbracelet/bubbletea"
 	"github.com/charmbracelet/lipgloss"
+	"github.com/muesli/reflow/wordwrap"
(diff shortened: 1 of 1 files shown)


bashbunni (approved):

meowgorithm:
In order for this to be merged text will need to be to reflowed (that is, re-wrapped) on resize. We'll want to pay special attention to how both the active line and viewport position if preserved (as best as possible) post-resize/re-wrap.

Additionally, I'd prefer to wait to merge this feature until https://github.com/muesli/reflow/issues/43 is fixed upstream so as to mitigate duplicate reports on that issue in this repo.

caarlos0:
this should be fixed on main/v2-exp!

thanks for the PR though!

PR #89, opened 2021-12-31
fix textinput infinite loop and panic

IllusionMan1212:
* fixes infinite loop when deleting input that only contains whitespace using deleteWordLeft()
* fixes index out of range panic when deleting input that only contains whitespace using deleteWordRight()


paralin (approved):

meowgorithm:
Well done. Thanks for this!

paralin:
No problem!

PR #90, opened 2022-01-03
fix(list): update keybindings when setting items

paralin:
Bug: when calling SetItems when items was previously empty, the keybindings for
up/down do not appear.

Fix: call updateKeybindings in SetItems.

### DIFF
--- a/list/list.go
+++ b/list/list.go
@@ -606,7 +606,7 @@ func (m *Model) updateKeybindings() {
 		m.KeyMap.CloseFullHelp.SetEnabled(false)
 
 	default:
-		hasItems := m.items != nil
+		hasItems := len(m.items) != 0
 		m.KeyMap.CursorUp.SetEnabled(hasItems)
 		m.KeyMap.CursorDown.SetEnabled(hasItems)
 
--- a/list/list.go
+++ b/list/list.go
@@ -292,6 +292,7 @@ func (m *Model) SetItems(i []Item) tea.Cmd {
 	}
 
 	m.updatePagination()
+	m.updateKeybindings()
 	return cmd
 }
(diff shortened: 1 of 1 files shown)


list/list.go
```diff
@@ -606,7 +607,7 @@ func (m *Model) updateKeybindings() {
 		m.KeyMap.CloseFullHelp.SetEnabled(false)
 
 	default:
-		hasItems := m.items != nil
+		hasItems := len(m.items) != 0
```
meowgorithm, line 610: These two statements are actually identical. I believe in this library we’re using the shorter convention.

meowgorithm:
Good catch! This generally looks right. We’ll merge once we verify. Thanks for the PR!

list/list.go
paralin: It's not identical. A nil slice is not the same thing as a slice with zero elements. The zero-element slice will return true for the != nil check.

paralin: You can see a proof of this here:

https://go.dev/play/p/0hlHLaxSACV

meowgorithm: Good call: I stand corrected

paralin:
@meowgorithm after extensive testing I can confidently say this was the fix.

meowgorithm:
Thanks so much, @paralin. Would it be possible to provide any sample code we could verify with over here? This will help us accelerate getting this ticket merged.

paralin:
@meowgorithm just create a bubbles list of length 0, then call AddItem to add an item (in Update).

meowgorithm:
Thanks so much for the fix!

paralin:
Thank you!

PR #98, opened 2022-01-16
add helper method to check if list is filtered

jon4hz:
I've added a helper method `IsFiltered()` to easily check whether a list is filtered or not.


maaslalani:
Seems very reasonable! Thank you @jon4hz!

PR #101, opened 2022-01-19
fix(textinput): use old cursor pos and simplify logic

IllusionMan1212:


meowgorithm:
Note: this fixes a regression introduced by https://github.com/charmbracelet/bubbles/commit/764fd321f617e8c57605a1d3919324b534780fea.

PR #104, opened 2022-01-22
list: fix to not show up quit in help while filtering

lusingander:
It’s correct that can't `quit` by pressing q when filtering, so it seems better to not show `q` in the help while filtering.


maaslalani:
Thank you! This seems correct, it seems though we were simply hiding the full help so it didn't seem to show up on some of the examples.

PR #108, opened 2022-01-26
fix(list): DisableQuitKeybinding is ignored after updating the list

aymanbagabas:
Don't set the quit keybinding in `updateKeybindings`


list/list.go
```diff
@@ -602,7 +602,6 @@ func (m *Model) updateKeybindings() {
 		m.KeyMap.ClearFilter.SetEnabled(false)
 		m.KeyMap.CancelWhileFiltering.SetEnabled(true)
 		m.KeyMap.AcceptWhileFiltering.SetEnabled(m.FilterInput.Value() != "")
-		m.KeyMap.Quit.SetEnabled(true)
```
muesli, line 605: Shouldn't this be here to disable the `Quit` key binding during filtering? (calling it with `false` instead of `true`)

aymanbagabas: This will disable the quit keymap, when it's enabled, after filtering. A better approach could be to store the keymap enabled value, disable it when "Filtering", and set it again after.

PR #114, opened 2022-02-02
feat: added `Validate` function for textinput

IllusionMan1212:
This PR adds a `Validate` function to the `textinput` bubble to allow for input validation before it's appended to the `textinput` value. 

3 built-in validators are provided:
- `ValidateInt`: validates integer input only.
- `ValidateFloat`: validates float input.
- `ValidateBool`: validates boolean-like inputs such as `y`, `n`, `t`, `f`, `1`, `0` (case insensitive)

A usage example is provided here: https://gist.github.com/IllusionMan1212/acddb1533790ad6abc6ee23fac6d9709

~~**This does not validate pasted input.** But should it?~~

### DIFF
--- a/textinput/textinput.go
+++ b/textinput/textinput.go
@@ -2,6 +2,7 @@ package textinput
 
 import (
 	"context"
+	"strconv"
 	"strings"
 	"sync"
 	"time"
@@ -91,6 +92,32 @@ func (c CursorMode) String() string {
 	}[c]
 }
 
+// ValidateBool validates boolean-like inputs only
+// i.e 'y', 'n', 't', 'f', '1', '0' (case insensitive).
+func ValidateBool(s string, m Model) error {
+	_, parseErr := strconv.ParseBool(string(s))
+
+	if parseErr == nil || s == "y" || s == "Y" || s == "n" || s == "N" {
+		return nil
+	} else {
+		return parseErr
+	}
+}
+
+// ValidateInt validates integer-only inputs.
+func ValidateInt(s string, m Model) error {
+	_, parseErr := strconv.ParseInt(s, 10, 64)
+
+	return parseErr
+}
+
+// ValidateFloat validates float-only inputs.
+func ValidateFloat(s string, m Model) error {
+	_, parseErr := strconv.ParseFloat(s, 64)
+
+	return parseErr
+}
+
 // Model is the Bubble Tea model for this text input element.
 type Model struct {
 	Err error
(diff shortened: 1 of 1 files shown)


textinput/textinput.go
```diff
@@ -15,6 +16,9 @@ import (
 
 const defaultBlinkSpeed = time.Millisecond * 530
 
+var intRegex = regexp.MustCompile("[0-9]")
```
muesli, line 19: Maybe we could avoid using regexps by relying on `strconv.ParseInt`, `strconv.ParseFloat` and `strconv.ParseBool`?

IllusionMan1212: done

textinput/textinput.go
```diff
@@ -91,6 +92,66 @@ func (c CursorMode) String() string {
...
+			out = append(out, r)
+		}
+	}
+
+	return out
+}
+
+// ValidateFloat validates float-only inputs.
+func ValidateFloat(runes []rune, m Model) []rune {
+	out := make([]rune, 0)
```
muesli, line 127: Any issue with using https://pkg.go.dev/strconv#ParseFloat here as well?

textinput/textinput.go
```diff
@@ -91,6 +92,66 @@ func (c CursorMode) String() string {
...
+	}
+
+	return out
+}
+
+// ValidateInt validates integer-only inputs.
+func ValidateInt(runes []rune, m Model) []rune {
+	out := make([]rune, 0)
+	for _, r := range runes {
+		_, parseErr := strconv.ParseInt(string(r), 10, 32)
```
muesli, line 116: I'd probably default to 64 bit here.

textinput/textinput.go
IllusionMan1212: I'm not sure if there's any advantage to using that as opposed to ParseInt since we're parsing a single digit every time

textinput/textinput.go
IllusionMan1212: Again, any argument why since we're parsing a single digit?

textinput/textinput.go
muesli: I see, I read that change wrong. I would pass the entire content of the `textinput` to the Validate method whenever its content changes. There are too many different ways of manipulating its content, I fear the code is becoming too convoluted quickly: entering characters, entering characters after moving the cursor, the clipboard, changing it with `textinput`'s API.

muesli:
We should also think of a way to notify API-users of validation issues, so they can inform the user of an error in their program.

textinput/textinput.go
IllusionMan1212: Right. I'm not sure if refactoring that is within the scope of this PR, but I could give it a try and open a separate PR before having this merged

meowgorithm:
Nice work so far. Just chiming in that we should definitely validate pasted input too.

IllusionMan1212:
Updated to accommodate the requested changes

PR #119, opened 2022-02-16
Add note to viewport that width/height are a no-op

BodeSpezial:
Because I just spent a few hours debugging why I was not able to set the width of the viewport I propose a note in the documentation to prevent similar confusion in the future. :woozy_face:

### DIFF
--- a/README.md
+++ b/README.md
@@ -86,7 +86,7 @@ indenting and text wrapping.
 
 ### Note:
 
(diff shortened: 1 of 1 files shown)


maaslalani:
Hey there! I don't think the `README.md` is the correct place for this information as it is mainly for demos and I doubt people would look in the `README.md` (in our case) for technical information about the component.
I'd suggest adding this information in the comment above the `Style lipgloss.Style` property of the component: https://github.com/charmbracelet/bubbles/blob/7cc578698457093a7ab8b2f293852acd52d7a8a9/viewport/viewport.go#L41-L43

Also, I'm sorry that you had to spend a few hours debugging 😞 but thank you so much for the PR to potentially save other devs time if they run into the same issue ❤️

PR #122, opened 2022-03-01
Add Blink method to textinput to return blink state

aschey:
As discussed on Slack earlier, this PR adds a `Blink()` method that returns the text input's blink state. This is useful if you want to render something more complex than the existing `View()` method can handle. For context, I'm implementing an input with multiple types of placeholders, each with different styles. I thought about adding something to the text input itself to handle this case, but it would add a lot of complexity for a rather niche use case. With the `blink` property exposed, users can implement their own `View()` method.

### DIFF
--- a/textinput/textinput.go
+++ b/textinput/textinput.go
@@ -203,6 +203,11 @@ func (m Model) Cursor() int {
 	return m.pos
 }
 
+// Blink returns whether or not to draw the cursor.
+func (m Model) Blink() bool {
+	return m.blink
+}
+
 // SetCursor moves the cursor to the given position. If the position is
 // out of bounds the cursor will be moved to the start or end accordingly.
 func (m *Model) SetCursor(pos int) {


textinput/textinput.go
```diff
@@ -203,6 +203,11 @@ func (m Model) Cursor() int {
 	return m.pos
 }
 
+// Blink returns whether the cursor should blink or not.
```
meowgorithm, line 206: I'd reword this to “Blink returns whether or not to draw the cursor,” just to disambiguate it from cursor states like `CursorStatic`, `CursorBlink` and `CursorHide`.

meowgorithm:
One small comment above, otherwise we're good to go. Thanks for this one!

textinput/textinput.go
aschey: Thanks, I agree that sounds much better. Fixed now.

meowgorithm (approved):

PR #123, opened 2022-03-02
fix(list): rendering empty list

aymanbagabas:
Return "No items found." when list is empty

Fixes: 4e18245481ab ("Add list bubble")


meowgorithm (approved):

meowgorithm:
Very nice catch!

PR #131, opened 2022-03-29
Don't render anything in Titlebar when everything is disabled

### LINKED ISSUE #130: List: Gap for Title/Spinner always present, even though both is disabled
List will always keep this gap open, even if `Title` and `Spinner` are disabled.

Bildschirmfoto 2022-03-28 um 22 39 46

That makes it less flexible to use. It should be able to appear completely compact, if you choose to. 

- [ ] Don't reserve space if nothing (title or spinner) are shown here
- [ ] Preferably: Add a `SetShowSpinner`-func that lets you disable it

toadle:
Probably fixes #130

### DIFF
--- a/spinner/spinner.go
+++ b/spinner/spinner.go
@@ -1,13 +1,11 @@
 package spinner
 
 import (
-	"strings"
 	"sync"
 	"time"
 
 	tea "github.com/charmbracelet/bubbletea"
 	"github.com/charmbracelet/lipgloss"
-	"github.com/muesli/reflow/ansi"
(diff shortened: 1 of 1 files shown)


meowgorithm:
Oh that's clever!

toadle:
I just flunked around styling a bit and noticed the Padding-gap that stayed when nothing was within view.
Dunno if this `len(view)` check is a good thing, you probably know smarter is. Anyway I fixed the linter issue and this helps me with my list styling.

PR #133, opened 2022-03-30
Add AppendContent

armlesshobo:
I wanted the ability to append content to the viewport without maintaining content externally and resetting the content every time the content changed.

What I ended up doing was taking the SetContent logic and changing it up so that we don't replace the internal model string slice, but instead append to it. This implies having the string split up into lines  first and then appended to the model. This method still respects YOffset and moves to the bottom if the offset is greater than the number of lines.

SetContent now functions by setting m.lines to nil, and then calling AppendContent.

### DIFF
--- a/viewport/viewport.go
+++ b/viewport/viewport.go
@@ -100,8 +100,14 @@ func (m Model) ScrollPercent() float64 {
 // SetContent set the pager's text content. For high performance rendering the
 // Sync command should also be called.
 func (m *Model) SetContent(s string) {
+	m.lines = nil
+	m.AppendContent(s)
+}
+
+// AppendContent append a string to the pager's text content. For high performance rendering the Sync command should also be called.
+func (m *Model) AppendContent(s string) {
 	s = strings.ReplaceAll(s, "\r\n", "\n") // normalize line endings
-	m.lines = strings.Split(s, "\n")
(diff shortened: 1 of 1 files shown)


armlesshobo:
bump

meowgorithm:
Thanks for the PR. The functionality makes sense, and your implementation seems good, though I have a couple concerns.

The first is that this will always append to a new line. Consider the following:

```go
vp := viewport.New()
vp.SetContent("I love cat")
vp.AppendContent("s a lot!!")
vp.View()

// Outputs:
// I love cat
// s a lot!!"
```

But it would be easy enough to adjust the behavior so the above would work, as well as add an additional method for your use case that potentially works more or less like the following:

```go
vp.AppendLines("Here I am appending a lines.\nAnd here is another.")
vp.AppendLines("Here's another line", "And here's one more")
```

My second, larger concern it that, at the moment, the burden of line wrapping is left to the user, and this becomes particularly important when terminal resizes come into play. When the content is managed internally there is no longer a way to re-run line wrapping.

However perhaps it makes sense to finally move that logic internally (see #56) which would mean removing the code that explicitly disallows it on the main `lipgloss.Style` member [here](https://github.com/charmbracelet/bubbles/blob/master/viewport/viewport.go#L361-L364). If we do decide this makes sense we should do that in a separate PR and merge it prior to this one.

PR #134, opened 2022-03-31
feat(progress): add minimum percent change needed to trigger an animation in progress

meowgorithm:
This update adds a threshold which prevents progress meters from animating over very short distances. Because terminal cells are rather large units, it's often a better experience to have short progress bar increments happen instantly.


caarlos0:
👀

PR #138, opened 2022-04-10
Fix timer Stop

padilo:
I don't know the details, but when I used Stop function on timer it doesn't stop the timer, but it worked fine if I use Toggle() function instead on a running Timer.

Looking at the code this sounds as it should be this way, but I don't know if I'm doing something wrong os this piece of code should be fixed.


maaslalani:
This totally looks correct to me. Thanks so much!

PR #148, opened 2022-05-03
New spinner with option

inkel:
This PR introduces a `spinner.Option` type that allows passing configuration options when creating new spinners using `spinner.New`. This doesn't break existing code as it uses variadic arguments, so any existing code as the following should continue to work:

```go
s := spinner.New()
s.spinner = spinner.Dot
```

This code now can be written as:

```go
s := spinner.New(spinner.WithSpinner(spinner.Dot))
```

In order not to be overly verbose, a `With` function was defined for each of the pre-existing spinners, so one can have for instance:

```go
s := spinner.New(spinner.withPulse())
```

Last but not least it also adds a `WithStyle(lipgloss.Style)` option function.

This is mostly syntactic sugar, but helps when you have a custom model and want to create it:

```go
func New() MyModel {
    return MyModel{
        spinner: spinner.New(spinner.WithDot(), spinner.WithStyle(myStyle)),
    }
}
```

Without these functions the call above has to be written as:

```go
func New() MyModel {
    m := MyModel{
        spinner: spinner.New(),
    }
    m.spinner.Spinner = spinner.Dot
    m.spinner.Style = myStyle

    return m
}
```

### DIFF
--- a/spinner/spinner.go
+++ b/spinner/spinner.go
@@ -198,41 +198,14 @@ func Tick() tea.Msg {
 //
 type Option func(*Model)
 
-// WithSpinner sets the spinner
+// WithSpinner is an option to set the spinner.
 func WithSpinner(spinner Spinner) Option {
 	return func(m *Model) {
 		m.Spinner = spinner
 	}
 }
 
-// WithLine sets a Line spinner
-func WithLine() Option { return WithSpinner(Line) }
-
-// WithDot sets a Dot spinner
-func WithDot() Option { return WithSpinner(Dot) }
-
-// WithMiniDot sets a MiniDot spinner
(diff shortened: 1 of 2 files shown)


inkel:
Any updates on this? It'd be nice to at least have the CI checks feedback 😬

muesli:
We'll review this as soon as possible, just a bit more patience, please :slightly_smiling_face:

maaslalani:
Hey @inkel, this looks great. Thank you for adding docs/comments and tests!
I removed the aliases for the specific spinners since I personally didn't feel they were necessary as they increase the API surface area and we can always add them later if people request them but they would be difficult to take them out.

inkel:
Thank you @maaslalani !!! Your decision of removing the specific spinners makes total sense, good call!

PR #155, opened 2022-05-24
feat: allow to set the height of the item

### LINKED ISSUE #117: List: text appears and disappears with certain items
Including multi-line items with empty spaces and lines causes the list component to behave oddly. Lines appear and disappear as you move the selection up and down. 

## Steps to reproduce with library example

1. Take the [default list example](https://raw.githubusercontent.com/charmbracelet/bubbletea/66cea095eb4b3d5f92f7a2131c084dcf8e785e87/examples/list-default/main.go)
2. Add this item at the beginning of the slice in `main()`:

    ```
item{title: "Wait a moment", desc: `This item here
	Has indentation

	Many lines
	Some of them blank

	This causes text to appear
		and disappear as the cursor moves up
	and down
`}, ```
3. Move a few places down the list, then up again. It's impossible to miss.

caarlos0:
The user might want to show more than 2 lines, and, right now, if they
try to do so, things break.

This also ensures the item never takes more than the allowed height, preventing issues like #117

fixes #117


list/defaultitem.go
```diff
@@ -139,11 +149,22 @@ func (d DefaultDelegate) Render(w io.Writer, m Model, index int, item Item) {
...
 
+	if d.ShowDescription {
+		lines := strings.Split(desc, "\n")
+		if len(lines) > d.height-1 {
+			desc = strings.Join(lines[0:d.height-1], "\n")
+		}
+	}
+
 	// Prevent text from exceeding list width
 	if m.width > 0 {
```
caarlos0, line 160: can we assume this is always >0, if so I can move the if above inside this one and avoid one strings.Split :P

muesli:
LGTM. The original idea was to have the users implement their own Delegate should they need a more intricate (e.g. multi-line) view that goes beyond a single title and a description. Most people seem to get away with our `DefaultDelegate` however, so I guess this is a welcome change.

aymanbagabas (approved):

PR #164, opened 2022-06-02
Feature: Multi-line Text Input

### LINKED ISSUE #63: Multiline Textinput
I want to use the text input bubble to allow the user to enter some notes. They can enter newlines in the notes if they want. It looks like the text input component ignores the Enter key event. Is it possible to add multi-line support so that <Enter> gets translated to "\n"? Or is there a more appropriate component for this?

maaslalani:
Fixes #63 

multiline-textinput

Allows the `textinput` bubble to be used as a multi-line text input by specifying `Height` and `LineLimit` as properties on the bubble model.

### API Changes

* Adds `Height` property which specifies the display height of the multi-line text input.
* Adds `LineLimit` property which specifies the maximum number of lines the text input can handle.

> Note: in the current state of the `Height` and `LineLimit` should be equal (there is no vertical scrolling currently)

### Internal Logic Changes
* `value` is now a 2D array of runes (`[][]rune`) to account for multi-line input.
* `pos` is now `row` and `col` to account for multi-line input.
* `View` is now separated into `singleLineView` and `multilineView`

### DIFF
--- a/textarea/textarea.go
+++ b/textarea/textarea.go
@@ -415,8 +415,15 @@ func (m *Model) handleOverflow() {
 		for j := 0; j < len(l); j++ {
 			w += rw.RuneWidth(l[j])
 			if w >= m.Width {
-				overflow := l[j:]
-				m.value[i] = l[:j]
+				// We've hit the maximum number of characters we can allow on this line
+				// Let's work backwards until we find a nice break point
+				bp := j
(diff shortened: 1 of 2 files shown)


maaslalani:
Superseded by https://github.com/charmbracelet/bubbles/pull/166

PR #166, opened 2022-06-07
Feature: Textarea

maaslalani:
Fixes: #63
Adds `textarea` for multi-line text input.

### Features
* Word-wrapping
* Vertical scrolling
* Optional Line Numbers + Line Number Styling
* Optional Cursor Line + Cursor Line Styling


maaslalani:
Preview: https://stuff.charm.sh/bubbles-examples/textarea.gif


textarea-example

caarlos0:
CleanShot 2022-06-10 at 15 59 11@2x


last line went 1 char more than the previous ones 🤔

caarlos0:
If I set width > than the screen width, text is "hidden":

image

caarlos0:
If I set height to a high value (e.g. >10000), it only gets 10 lines and looks a weird:

image

textarea/textarea.go
```diff
@@ -0,0 +1,1246 @@
...
+	"unicode"
+
+	"github.com/atotto/clipboard"
+	"github.com/charmbracelet/bubbles/viewport"
+	tea "github.com/charmbracelet/bubbletea"
+	"github.com/charmbracelet/lipgloss"
+	rw "github.com/mattn/go-runewidth"
+)
+
+const defaultBlinkSpeed = time.Millisecond * 530
```
caarlos0, line 18: I almost wonder if we should, somehow, move the cursor into its own bubble?

It seems like it would be the same thing for both textarea and textinput

PR #169, opened 2022-06-10
feat: custom status bar title

### LINKED ISSUE #116: list: personalization of status bar message
Currently the status bar presents a fixed message with `X items` as highlighted in the image
Selezione_020

It could be interesting the possibility to customize the label `item(s)`

Here is the reference in the code
https://github.com/charmbracelet/bubbles/blob/18d25458daedd7508e55de8d64c55c1a8a6a677b/list/list.go#L1023

What do you think about ? (I can provide a PR about this)

wesleimp:
Closes #116

### DIFF
--- a/list/list.go
+++ b/list/list.go
@@ -1083,7 +1083,7 @@ func (m Model) statusView() string {
 		}
 	} else if len(m.items) == 0 {
 		// Not filtering: no items.
-		status = m.Styles.StatusEmpty.Render("No items")
+		status = m.Styles.StatusEmpty.Render("No " + m.itemNamePlural)
 	} else {
 		// Normal
 		filtered := m.FilterState() == FilterApplied
@@ -1138,7 +1138,7 @@ func (m Model) populatedView() string {
 		if m.filterState == Filtering {
 			return ""
 		}
-		return m.Styles.NoItems.Render("No items found.")
+		return m.Styles.NoItems.Render("No " + m.itemNamePlural + " found.")
 	}
 
 	if len(items) > 0 {
--- a/list/list.go
+++ b/list/list.go
@@ -298,7 +298,7 @@ func (m *Model) SetStatusBarItemName(singular, plural string) {
 	m.itemNamePlural = plural
 }
 
-// StatusBarItemName returns the custom status bar title
+// StatusBarItemName returns singular and plural status bar item names
 func (m Model) StatusBarItemName() (string, string) {
 	return m.itemNameSingular, m.itemNamePlural
 }
--- a/list/list.go
+++ b/list/list.go
@@ -287,13 +287,13 @@ func (m Model) ShowStatusBar() bool {
 	return m.showStatusBar
 }
 
-// SetStatusBarTitle shows or hides the view that displays metadata about the
-// list, such as item counts.
+// SetStatusBarTitle defines a replacement for the items identifier. Defaults
+// to item/items
 func (m *Model) SetStatusBarTitle(v string) {
 	m.statusBarTitle = v
 }
 
-// ShowStatusBar returns whether or not the status bar is set to be rendered.
+// StatusBarTitle returns the custom status bar title
 func (m Model) StatusBarTitle() string {
 	return m.statusBarTitle
 }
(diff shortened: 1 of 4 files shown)


list/list.go
```diff
@@ -1048,17 +1060,24 @@ func (m Model) statusView() string {
...
+	var title string
+	if m.statusBarTitle != "" {
+		title = fmt.Sprintf("%d %s", visibleItems, m.statusBarTitle)
+	} else {
+		plural := ""
+		if visibleItems != 1 {
+			plural = "s"
+		}
+
+		title = fmt.Sprintf("%d item%s", visibleItems, plural)
```
maaslalani, line 1072: I think this `if m.statusBarTitle != ""` statement can go away if we set the default value of `m.statusBarTitle` to `item` in the `New` constructor and then allow people to change it with `SetStatusBarTitle`.

Like we do with other defaults, here:
https://github.com/charmbracelet/bubbles/blob/a62c69a3a309a98fa35c97be01b8959113a522b9/list/list.go#L200-L221

maaslalani: I'd suggest setting `statusBarTitle:     "item"` in the constructor and then simply use that value, whether it is the default or user customized.

```suggestion
	plural := ""
	if visibleItems != 1 {
		plural = "s"
	}

	title = fmt.Sprintf("%d %s%s", visibleItems, m.statusBarTitle, plural)
```

Also, I wonder if we can change the `statusBarTitle` name to something like `itemName` to show what is being filtered so that people use the singular form so we can handle the pluralization.

wesleimp: Adding this to `New` constructor, we cannot handle plurals anymore. What do you think about to change the default to `item(s)` or something like that?

maaslalani: Maybe we should have a `itemNameSingular` and `itemNamePlural` that should be set? I.e. if someone is listing people they should set `itemNameSingular = "person"` and `item = "people"` and then we can set the defaults to `item` and `items`? That way we can also handle pluralization for the custom settings as well by performing the logic when `visibleItems > 1`.

What do you think?
I'm also not opposed to doing `item(s)` as this keeps the changing less when filtering and then simply having `statusBarTitle` but I might need to ask the team about this.

wesleimp: > Maybe we should have a itemNameSingular and itemNamePlural that should be set? I.e. if someone is listing people they should set itemNameSingular = "person" and item = "people" and then we can set the defaults to item and items? That way we can also handle pluralization for the custom settings as well by performing the logic when visibleItems > 1.

I like this approach.

list/list_test.go
```diff
@@ -0,0 +1,60 @@
+package list
+
+import (
+	"fmt"
+	"io"
+	"testing"
+
+	tea "github.com/charmbracelet/bubbletea"
+	"github.com/stretchr/testify/assert"
```
maaslalani, line 9: Thank you so much for adding tests, though I think I'd like to avoid adding the `testify/assert` library.

We can pretty much accomplish the same thing with normal `if` statements since `statusView` is simply a string.

Before:
```go
assert.Contains(list.statusView(), "1 item")
```

After:
```go
if !strings.Contains(list.statusView(), "1 item") {
    t.Fatalf("Error: expected view to contain %s", "1 item")
}
```

wesleimp: I'll remove it for now. But I think that as we add more test cases, it's pretty much hard to maintain with a normal `if`.

maaslalani: Much appreciated ❤️

list/list.go
```diff
@@ -286,6 +291,18 @@ func (m Model) ShowStatusBar() bool {
...
 }
 
+// SetStatusBarItemName defines a replacement for the items identifier. Defaults
+// to item/items
+func (m *Model) SetStatusBarItemName(singular, plural string) {
+	m.itemNameSingular = singular
+	m.itemNamePlural = plural
+}
+
+// StatusBarItemName returns the custom status bar title
```
maaslalani, line 301: Can we add a note here about the return values, I want to make sure people know the order the names come back in
```suggestion
// StatusBarItemName returns singular and plural status bar item names
```

PR #181, opened 2022-06-28
Delegate Cursor Logic to `cursor` bubble

maaslalani:
This PR extracts the original cursor logic to the cursor bubble.


textinput/textinput.go
```diff
@@ -216,74 +136,26 @@ func (m Model) Value() string {
 	return string(m.value)
 }
 
-// Cursor returns the cursor position.
-func (m Model) Cursor() int {
+// Position returns the cursor position.
+func (m Model) Position() int {
```
maaslalani, line 140: ⚠️ Breaking change! Since I use the name `Cursor` for the `cursor.Model`.

PR #184, opened 2022-07-05
fix(textinput): set cursor on initial value

GabrielNagy:
This reverts the code back to the logic from 4ce16e8 which fixed the issue of the cursor not being moved when an initial value was set with SetValue.

The fix regressed in 16053f4.


maaslalani:
Thank you for the PR, I will test it out just to make sure the validation stuff continues to works correctly :)

maaslalani:
Looks great, thank you so much!

maaslalani (approved):

PR #185, opened 2022-07-05
feat(textinput): do not block input on validation

GabrielNagy:
This PR builds upon the excellent work in https://github.com/charmbracelet/bubbles/pull/167 and https://github.com/charmbracelet/bubbles/pull/114 and makes a breaking change to the validation API.

Currently, validation will completely block text input if the Validate function returns an error. This is now changed so the function no longer blocks input if this is the case, thus handing this responsibility to the clients.

This is helpful for cases where the user is requested to type an existing system path, and the `Validate` function keeps asserting the existence of the path. With the current implementation such a validation is not possible.

For example:

    > /
    Err: nil

    > /t
    Err: /t: No such file or directory

    > /tm
    Err: /tm: No such file or directory

    > /tmp
    Err: nil

### DIFF
--- a/textinput/textinput.go
+++ b/textinput/textinput.go
@@ -181,19 +181,14 @@ func (m *Model) SetValue(s string) {
 	// Clean up any special characters in the input provided by the
 	// caller. This avoids bugs due to e.g. tab characters and whatnot.
 	runes := m.san().Sanitize([]rune(s))
-	m.setValueInternal(runes)
+	err := m.validateIfDefined(string(runes))
+	m.setValueInternal(runes, err)
 }
 
-func (m *Model) setValueInternal(runes []rune) {
-	if m.Validate != nil {
-		if err := m.Validate(string(runes)); err != nil {
-			m.Err = err
-			return
-		}
-	}
+func (m *Model) setValueInternal(runes []rune, err error) {
+	m.Err = err
 
 	empty := len(m.value) == 0
-	m.Err = nil
 
 	if m.CharLimit > 0 && len(runes) > m.CharLimit {
 		m.value = runes[:m.CharLimit]
@@ -307,8 +302,6 @@ func (m *Model) insertRunesFromUserInput(v []rune) {
 	tail := make([]rune, len(tailSrc))
 	copy(tail, tailSrc)
 
-	oldPos := m.pos
-
 	// Insert pasted runes
 	for _, r := range paste {
 		head = append(head, r)
@@ -323,11 +316,8 @@ func (m *Model) insertRunesFromUserInput(v []rune) {
 
 	// Put it all back together
 	value := append(head, tail...)
-	m.setValueInternal(value)
-
-	if m.Err != nil {
-		m.pos = oldPos
-	}
+	inputErr := m.validateIfDefined(string(value))
+	m.setValueInternal(value, inputErr)
 }
 
 // If a max width is defined, perform some logic to treat the visible area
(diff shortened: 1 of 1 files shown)


maaslalani:
❤️ This is fantastic! @meowgorithm and I were _just_ talking about how this would be a great addition. Thank you for reading our minds @GabrielNagy 😄

maaslalani:
Hey @GabrielNagy sorry for forgetting about this one! I almost wonder if the `ValidateFunc` should return an error and a boolean on whether to block the input such that validation could block some input but not block all input (i.e. warnings vs errors). The difference would be you could then validate input and give warning errors such as "file not found" but also prevent input that will never be valid such as inserting a special character in a file name. This obviously would be a breaking API change so I might need to think about it a little more, but I'm curious to hear your thoughts on the validation function returning a bool for whether to allow the input or block it from changing. And, you can always return false or always return true for the behaviour in the current PR.

hopefulTex:
I really like the option to block or warn on invalid input!
Would it be reasonable to set a behavior flag, like `m.textInput.Warning()', and let the validator logic flip it back after writing the rune?
It shouldn't break any users' code, and they'd set the behavior flag before returning their `Err`.

As for the separate validation functions, would a more general `Submit()` function work for this? Maybe it could just run the former Validate function with `block` set before executing a user-supplied Submit function.
(I'd also love a `SubmitKey tea.Msg` to proc `Submit()` when set, but scope is scope.)

rynowak:
I just hit this issue as well, and I need to stop using the `Validate` function as a result. Is there anything we can do to accelerate getting this functionality added?

CoMfUcIoS:
Any updates on this? Willing to help if needed as well.

maaslalani:
@GabrielNagy, what do you think of simply removing the blocking of the validation?

That way we can keep the signature of the function the same. We would then simply put the responsibility on the user to ensure that there is no `m.Err` before they submit their form.

I like the idea of keeping the `func Validate(s string) err` since that the signature we use for `huh`, but for `huh` we just validate on `Blur`.

textinput/textinput.go
```diff
@@ -323,9 +322,10 @@ func (m *Model) insertRunesFromUserInput(v []rune) {
 
 	// Put it all back together
 	value := append(head, tail...)
-	m.setValueInternal(value)
+	blockInput, inputErr := m.validateIfDefined(string(value))
+	m.setValueInternal(value, inputErr, blockInput)
 
-	if m.Err != nil {
```
maaslalani, line 328: So essentially we just remove this line and never block.

maaslalani:
Thanks for all your patience on this one @GabrielNagy!

jippi:
👋 Thanks for looking at this - I just spent a couple of hours debugging this issue on my own project 😆

textinput/textinput.go
```diff
@@ -597,13 +591,12 @@ func (m Model) Update(msg tea.Msg) (Model, tea.Cmd) {
 			if m.pos < len(m.value) {
 				m.SetCursor(m.pos + 1)
 			}
-		case key.Matches(msg, m.KeyMap.DeleteWordBackward):
-			m.deleteWordBackward()
```
maaslalani, line 601: Any reason for deleting this?

maaslalani: Ah I see it's a duplicate now.

maaslalani (approved):
Awesome work @GabrielNagy, just tested and it works great!

PR #188, opened 2022-07-06
Unbound keys should not be enabled

motemen:
[Unbind()](https://pkg.go.dev/github.com/charmbracelet/bubbles@v0.13.0/key#Binding.Unbind)'ed keys are not usable anymore, so it should not be counted as Enabled. For example, unbound keys should not show up in help.


meowgorithm:
Good catch, and thank you for the contribution!

motemen:
Thank you for this quick merge! 🙌

PR #204, opened 2022-08-05
LineCount + Include New Line characters in Length()

### LINKED ISSUE #202: textarea: .Length doesn't properly include newline characters
The code in the textarea `Length` method does not match the definition in `Value`: it does not include the newline characters.

So if the text area contains, say, 2 lines of 4 characters each, Length will return 8 but the Value has length 9.

### LINKED ISSUE #203: textarea: need method to retrieve the current number of lines in the Value
It's already possible to count the number of newline characters in `Value`, however it's rather inefficient.

It would be better to have a constant-time accessor for the nr of lines in the input  (`ValueLines`?)

As well as an accessor for the nr of lines displayed that are part of the input, i.e. the sum of the LineInfo heights (I think?)

maaslalani:
Fixes #203
Fixes #202

### Changes
- fix: include new line characters for Length()
- feat: LineCount method for counting number of lines in textarea

### DIFF
--- a/textarea/textarea.go
+++ b/textarea/textarea.go
@@ -289,6 +289,11 @@ func (m *Model) Length() int {
 	return l + len(m.value) - 1
 }
 
+// LineCount returns the number of lines that are currently in the text input.
+func (m *Model) LineCount() int {
+	return len(m.value)
+}
+
 // Line returns the line position.
 func (m Model) Line() int {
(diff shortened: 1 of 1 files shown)


maaslalani:
Hey @knz, this PR should solve both of your issues 🙂

PR #205, opened 2022-08-06
Ability to add items to list

Vinschni:
I added the ability to add items using the filter prompt when pressing the "a" keys.

Furthermore for the textarea i added a function to explicitly set the line number.

### DIFF
--- a/list/list.go
+++ b/list/list.go
@@ -101,6 +101,9 @@ func DefaultFilter(term string, targets []string) []Rank {
 
 type statusMessageTimeoutMsg struct{}
 
+// Message when a new item is added with used entered value
+type ItemAddedMsg string
+
 // FilterState describes the current filtering state on the model.
 type FilterState int
 
@@ -109,6 +112,7 @@ const (
 	Unfiltered    FilterState = iota // no filter set
 	Filtering                        // user is actively setting a filter
 	FilterApplied                    // a filter is applied and user is not editing filter
+	AddingItem                       // User is adding a new item and entering the title
 )
 
 // String returns a human-readable string of the current filter state.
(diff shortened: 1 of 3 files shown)


bashbunni:
I think it makes sense to add this feature directly to the list... I'm trying to think of cases where a list might be used, but the program author would want the list to be immutable for the user 🤔

muesli:
While the code is fine, I think prepending a `textinput` can already easily be done as a `list` user. I agree that the convenience would be nice, but we should probably strive for a solution that lets you edit & add items inline (see glow for an example).

muesli:
> I think it makes sense to add this feature directly to the list... I'm trying to think of cases where a list might be used, but the program author would want the list to be immutable for the user thinking

I think there are plenty of such use cases. A fixed list of items isn't uncommon, like retrieving items from an API that doesn't just let you create new ones.

list/list.go
```diff
@@ -878,6 +922,12 @@ func (m *Model) handleFiltering(msg tea.Msg) tea.Cmd {
 			if m.FilterInput.Value() == "" {
 				m.resetFiltering()
 			}
+		case key.Matches(msg, m.KeyMap.AcceptWhileAdding):
+			m.hideStatusMessage()
+			m.FilterInput.Blur()
+			m.filterState = Unfiltered
+			m.updateKeybindings()
+			return func() tea.Msg { return ItemAddedMsg(m.FilterInput.Value()) }
```
meowgorithm, line 930: Small note here: logically speaking, there's no need to send a message for internal app communication. That is, commands are strictly for IO.

bashbunni:
@muesli 
Just saw this PR again, here are my thoughts. It sounds like we're leaning away from implementing this based on our discussion above. Should I close the ticket?

maaslalani:
Hey @Vinschni, thanks so much for this contribution! ❤️ 

I'm cleaning up some old PRs and it looks like we want a more generic / flexible solution for adding items. If you still need this feature please feel free to re-open this PR or create a new issue we can tackle!

PR #208, opened 2022-08-12
Avoid recomputing the key representation in Matches

knz:
cc @maaslalani


maaslalani:
Thanks @knz! Appreciate this change ❤️

PR #209, opened 2022-08-12
feat: added new spinners

BlueFalconHD:


meowgorithm:
Thanks for the PR. Unfortunately, we'd prefer to avoid features that require a specific font to be installed (such as the 'languages' spinner in this case).

That said, it's totally doable to define custom spinners in your application and use them as needed.

BlueFalconHD:
> Thanks for the PR. Unfortunately, we'd prefer to avoid features that require a specific font to be installed (such as the 'languages' spinner in this case).
> 
> That said, it's totally doable to define custom spinners in your application.

It's all good! Just trying to contribute to more projects. I assume the same thing goes for emoji because those depend on fonts as well.

PR #210, opened 2022-08-14
textarea: misc improvements

knz:
I'm working to make textarea closer to readline/editline conventions (e.g. as used on a shell command line or emacs).

- search prev/next word should find words on the prev/next line if there's no word remaining on the current line.
- Alt+u to uppercase cur/next word
- Alt+l to lowercase cur/next word
- Alt+c to capitalize cur/next word
- Ctrl+t to transpose previous 2 characters

cc @maaslalani

### DIFF
--- a/textarea/textarea.go
+++ b/textarea/textarea.go
@@ -46,6 +46,10 @@ type KeyMap struct {
 	Paste                   key.Binding
 	WordBackward            key.Binding
 	WordForward             key.Binding
+
+	UppercaseWordForward  key.Binding
+	LowercaseWordForward  key.Binding
+	CapitalizeWordForward key.Binding
 }
 
 // DefaultKeyMap is the default set of key bindings for navigating and acting
@@ -67,6 +71,10 @@ var DefaultKeyMap = KeyMap{
 	LineStart:               key.NewBinding(key.WithKeys("home", "ctrl+a")),
 	LineEnd:                 key.NewBinding(key.WithKeys("end", "ctrl+e")),
 	Paste:                   key.NewBinding(key.WithKeys("ctrl+v")),
+
+	CapitalizeWordForward: key.NewBinding(key.WithKeys("alt+c")),
+	LowercaseWordForward:  key.NewBinding(key.WithKeys("alt+l")),
+	UppercaseWordForward:  key.NewBinding(key.WithKeys("alt+u")),
 }
 
 // LineInfo is a helper for keeping track of line information regarding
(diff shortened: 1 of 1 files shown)


knz:
Using this in https://github.com/knz/bubbline

textarea/textarea.go
```diff
@@ -547,62 +577,103 @@ func (m *Model) deleteWordRight() {
...
+	m.doWordRight(func(bool, int) { /* nothing */ })
+}
 
-	i := m.col
-	for i < len(m.value[m.row]) {
-		if unicode.IsSpace(m.value[m.row][i]) {
-			m.SetCursor(m.col + 1)
-			i++
-		} else {
+func (m *Model) doWordRight(fn func(firstChar bool, pos int)) {
```
maaslalani, line 634: Instead of having `doWordRight` take a function with a `firstChar bool`, I'd suggest it take a index of the start of the word so that it is a bit more flexible for stuff like deleting words / referencing the start of the word index later while still being able to do `start == pos` to get whether the current character is the first character, what do you think?

```suggestion
func (m *Model) doWordRight(fn func(start int, pos int)) {
```

knz: Done, ready for another look.

maaslalani: Thanks @knz, this looks absolutely fantastic! Will merge soon after a quick review!

maaslalani:
@knz quick question (since I don't use emacs) is the <kbd>ctrl+t</kbd> transpose supposed to move the cursor forward or should it stay in the same place after the transpose.

textarea/textarea.go
```diff
@@ -547,62 +577,103 @@ func (m *Model) deleteWordRight() {
...
 
-	i := m.col
-	for i < len(m.value[m.row]) {
-		if unicode.IsSpace(m.value[m.row][i]) {
-			m.SetCursor(m.col + 1)
-			i++
-		} else {
+func (m *Model) doWordRight(fn func(charIdx int, pos int)) {
+	// Skip spaces forward.
+	for {
```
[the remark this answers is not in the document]
knz: Fixed.

knz:
> quick question (since I don't use emacs) is the ctrl+t transpose supposed to move the cursor forward or should it stay in the same place after the transpose.

It transposes the last two characters, then moves the cursor forward.

The point is that if your cursor is at this position:  `a.bcdef`

And then you press Ctrl+T continuously multiple times, you get the `a` to move to the end.

maaslalani:
Another question: If I am inside a word:

```
the qu|ick fox
```

and I press <kbd>alt+c</kbd> to capitalize the word,
should the result be:

```
the Quick fox
```

or

```
the quIck fox
```

maaslalani:
Just tested out in `emacs` looks like the implemented behaviour is correct (matches emacs) so this looks great! Thank you @knz for adding this change ❤️

PR #211, opened 2022-08-14
textarea: support dynamic prompts

knz:
The secondary prompt is displayed on every line after the first.
See for example:
[asciicast](https://asciinema.org/a/iFBPBwoDZOzcoRJOgfmPk8ObH)

Used in https://github.com/knz/bubbline

cc @maaslalani

### DIFF
--- a/textarea/textarea.go
+++ b/textarea/textarea.go
@@ -119,6 +119,11 @@ type Model struct {
 	// General settings.
 
 	// Prompt is printed at the beginning of each line.
+	//
+	// When changing the value of Prompt after the model has been
+	// initialized, ensure that SetWidth() gets called afterwards.
+	//
+	// See also SetPromptFunc().
 	Prompt string
 
 	// Placeholder is the text displayed when the user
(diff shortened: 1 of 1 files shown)


knz:
ping @maaslalani

maaslalani:
Hey @knz,

This change seems very specific for line editing, the `textarea` bubble is meant to emulate the HTML `textarea` so that people are familiar with it and it is easy to understand / use.

https://developer.mozilla.org/en-US/docs/Web/HTML/Element/textarea

I'll have to double check with the team, but it seems that this change is increasing the API surface area for a single use case which I would imagine most people are not going to use, which means it might be better handled in a more generic way. 

I can totally see why this is useful for a line editing application. One more generic implementation that comes to mind (not saying this is the correct approach) is to perhaps allow a `PromptFunc` to be defined that passes the line number and allows the prompt to be defined as a function of the line number that way it is at least a bit more generic i.e. people can implement alternating prompts for even or odd lines, or hide the prompt past line 5, or make the first and last prompts a special characters, and this next prompt logic could be implemented.

knz:
PromptFunc works for me!

knz:
i'll make the change

maaslalani:
Thanks @knz, much appreciated!

knz:
Hi @maaslalani I have changed the logic to use a dynamic prompt as you suggested!

maaslalani:
This looks absolutely fantastic @knz, thank you so much!

Another use case I just thought of for having the `PromptFunc` take a line number is to show diagnostics / error warnings for certain line numbers that have parse errors on them. So this change will be extremely useful for that!

PR #218, opened 2022-08-19
fix(help): full help renders fully when width is unset.

### LINKED ISSUE #217: Help in FullView doesn't render with 0 width
Unlike in ShorView, when the width is 0, FullView doesn't render anything. I believe that 0 should indicate unknown or automatic as it does for the ShortView.

craiggwilson:
Fixes #217


knz:
cc @meowgorithm

maaslalani:
Works great for me! Thank you for this PR.

PR #219, opened 2022-08-20
feat(table): add function to set cursor position

nikaro:
This would allow to (re)set the cursor position after the table is updated.

### DIFF
--- a/table/table.go
+++ b/table/table.go
@@ -291,6 +291,12 @@ func (m Model) Cursor() int {
 	return m.cursor
 }
 
+// SetCursor sets the cursor position in the table.
+func (m *Model) SetCursor(n int) {
+	m.cursor = clamp(n, 0, len(m.rows)-1)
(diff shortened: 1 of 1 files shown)


table/table.go
```diff
@@ -291,6 +291,12 @@ func (m Model) Cursor() int {
 	return m.cursor
 }
 
+// SetCursor sets the cursor position in the table.
+func (m *Model) SetCursor(n int) {
```
maaslalani, line 295: I wonder if `MoveTo` is a better name, to match `MoveUp` and `MoveDown`.

nikaro: I used SetCursor since it matches with `Cursor` and the others `Set*` methods, and it literally sets the `cursor` attribute :-) But it's up to you, i don't have a preference on this, let me know if you want me to amend the commit.

maaslalani (approved):
LGTM! Thanks @nikaro

PR #220, opened 2022-08-21
fix(table): correct keybinding for page down

buztard:
Fixes `pgdown` keybindings and adds keys to help.


maaslalani:
Thank you!

PR #226, opened 2022-08-29
textarea: new bindings for "go to begin" / "go to end"

knz:
Explanation about the chosen bindings:

- Alt+< / Alt+>: These are the same as emacs.
- Ctrl+Home / Ctrl+End: These are the same as some other windows editors.

(For the latter, they will only work after https://github.com/charmbracelet/bubbletea/pull/396 is merged. But this PR can merge ahead of it -- unrecognized bindings are simply ignored for the time being.)

cc @maaslalani

### DIFF
--- a/textarea/textarea.go
+++ b/textarea/textarea.go
@@ -46,6 +46,8 @@ type KeyMap struct {
 	Paste                   key.Binding
 	WordBackward            key.Binding
 	WordForward             key.Binding
+	InputBegin              key.Binding
+	InputEnd                key.Binding
 
 	UppercaseWordForward  key.Binding
 	LowercaseWordForward  key.Binding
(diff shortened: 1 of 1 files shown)


textarea/textarea.go
```diff
@@ -46,6 +46,8 @@ type KeyMap struct {
 	Paste                   key.Binding
 	WordBackward            key.Binding
 	WordForward             key.Binding
+	MoveToBegin             key.Binding
+	MoveToEnd               key.Binding
```
maaslalani, line 50: I would probably name this `InputStart` / `InputEnd` to match `LineStart` / `LineEnd`

maaslalani (approved):

textarea/textarea.go
knz: Sure, done.

maaslalani:
Thanks so much @knz! Much appreciated ❤️

PR #228, opened 2022-08-31
fix(viewport): properly truncate to size

### LINKED ISSUE #227: viewport: width and height are not respected if input contains long lines
As discussed here https://charmbracelet.slack.com/archives/CQ2E9CGTV/p1661964596034599?thread_ts=1661940898.978989&cid=CQ2E9CGTV

knz:
Fixes #227 

- logical content is wider than display width (long lines).
  In this case, we truncate.
- logical content is smaller than display height (fewer lines
  than visible).
  In this case, we pad with empty lines up to the specified height.
- logical content is higher than display height (more lines than
  visible).
  In this case, we truncate.

- style specifies a width wider than the display width.
  In this case, we ignore the style width and
  fit in the display width.
- style specifies a height higher than the display height.
  Same as width, we ignore the style and fit in the display height.
- style specifies a narrower width or smaller height than the display.
  In this case we obey the style.

cc @meowgorithm

### DIFF
--- a/viewport/viewport.go
+++ b/viewport/viewport.go
@@ -349,12 +349,24 @@ func (m Model) View() string {
 		// position anything below this view properly.
 		return strings.Repeat("\n", max(0, m.Height-1))
 	}
-	contentWidth := m.Width - m.Style.GetHorizontalFrameSize()
-	contentHeight := m.Height - m.Style.GetVerticalFrameSize()
-	return m.Style.Copy().
-		UnsetWidth().MaxWidth(contentWidth).            // truncate long lines.
-		Height(contentHeight).MaxHeight(contentHeight). // pad to height then truncate.
+
+	w, h := m.Width, m.Height
+	if sw := m.Style.GetWidth(); sw != 0 {
+		w = min(w, sw)
+	}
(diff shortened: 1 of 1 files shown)


meowgorithm:
Thanks for this @knz. One thing I'm noticing is that borders and padding doesn't seem to accounted for at render time (see [example](https://gist.github.com/meowgorithm/adee9aa8341aac0096b642c518aae949)). Otherwise, everything seems good.

knz:
Ok I've updated it. 

But note that your example here sets a width in the style, which none of the bubbles really support well (I'm fixing viewport in this PR, but the other bubbles will remain weird in that case).

This of course would benefit from unit tests. I'll send a separate PR for that.

meowgorithm:
Nicely done, @knz. I've played with it a bit and it seems perfect now.

> But note that your example here sets a width in the style, which none of the bubbles really support well (I'm fixing viewport in this PR, but the other bubbles will remain weird in that case).

If you're referring to [this part](https://gist.github.com/meowgorithm/adee9aa8341aac0096b642c518aae949#file-main-go-L39), it's only for text wrapping, which I imagine is an opt-in feature we should add at some point.

PR #230, opened 2022-09-03
fix(cursor): apply TextStyle inline

### LINKED ISSUE #229: cursor: TextStyle is not applied inline
When the cursor is blinking, the TextStyle is not applied inline. So if it contains padding / borders etc, mayhem ensues.

I have a PR for this.

cc @maaslalani

knz:
Found bug using [catwalk](https://github.com/knz/catwalk).

Fixes #229.

cc @maaslalani

### DIFF
--- a/cursor/cursor.go
+++ b/cursor/cursor.go
@@ -200,9 +200,8 @@ func (m *Model) SetChar(char string) {
 
 // View displays the cursor.
 func (m Model) View() string {
-	style := m.TextStyle.Inline(true)
-	if !m.Blink {
-		style = m.Style.Inline(true).Reverse(true)
+	if m.Blink {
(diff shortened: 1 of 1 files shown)


cursor/cursor.go
```diff
@@ -200,8 +200,9 @@ func (m *Model) SetChar(char string) {
 
 // View displays the cursor.
 func (m Model) View() string {
-	if m.Blink {
-		return m.TextStyle.Render(m.char)
+	style := m.TextStyle.Inline(true)
```
maaslalani, line 203: I don't think we really need to allocate a variable here (to be fair, I'd guess the Go compiler does optimize this for us), we can use the same return logic as before and add an `Inline(true)` to the style.

PR #250, opened 2022-09-25
fix: textarea paste

hhe07:
while trying to put together a program using this bubble, i discovered that the paste function seems to be incomplete (at least, compared to the ``textinput`` paste function), in that it doesn't apply changes to ``textarea.value``.

i just copied the associated code from ``textinput`` to actually apply the changes, it's nothing special.

### DIFF
--- a/textarea/textarea.go
+++ b/textarea/textarea.go
@@ -494,6 +494,10 @@ func (m *Model) handlePaste(v string) {
 		}
 	}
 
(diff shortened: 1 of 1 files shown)


maaslalani:
Thank you @hhe07, taking a look now!

maaslalani:
Works great, thanks you so much for the PR!

PR #261, opened 2022-10-02
key: optimize `Matches`

knz:
Fixes #475.
```
name                old time/op    new time/op    delta
Matches/success-32    66.9ns ± 1%    15.7ns ± 0%   -76.47%  (p=0.016 n=5+4)
Matches/fail-32       27.9ns ± 0%    12.4ns ± 2%   -55.70%  (p=0.016 n=4+5)

name                old alloc/op   new alloc/op   delta
Matches/success-32     5.00B ± 0%     0.00B       -100.00%  (p=0.008 n=5+5)
Matches/fail-32        0.00B          0.00B           ~     (all equal)

name                old allocs/op  new allocs/op  delta
Matches/success-32      1.00 ± 0%      0.00       -100.00%  (p=0.008 n=5+5)
Matches/fail-32         0.00           0.00           ~     (all equal)
```

cc @meowgorithm @muesli

### DIFF
--- a/key/key.go
+++ b/key/key.go
@@ -37,13 +37,15 @@
 package key
 
 import (
+	"strings"
+
 	tea "github.com/charmbracelet/bubbletea"
 )
 
 // Binding describes a set of keybindings and, optionally, their associated
 // help text.
 type Binding struct {
-	keys     []string
+	keys     []tea.Key
 	help     Help
 	disabled bool
 }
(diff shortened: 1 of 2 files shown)


knz:
NB: I could imagine the map from string to tea.Key being defined in the bubbletea package directly. If you prefer that, let me knpw.

muesli:
> NB: I could imagine the map from string to tea.Key being defined in the bubbletea package directly. If you prefer that, let me know.

I think that makes sense and is probably nicer next to the other `tea.Key` code.

maaslalani:
Whoa! This is awesome @knz, solid optimization!

knz:
> I think that makes sense and is probably nicer next to the other tea.Key code.

Ok see https://github.com/charmbracelet/bubbletea/pull/491

Then to merge this PR I'll need a release number for the go.mod update.

maaslalani:
@muesli Is this good to merge?

PR #265, opened 2022-10-05
Improved performance, added individual cell toggle, and move column right/left

mathaou:
I ended up not adding mouse support since that is only easy when in alt mode, so I'll add altmode/mouse support in a separate PR. This one is mainly to support arbitrarily large .csv files with thousands of rows and columns.


mathaou:
I realized I need to update for runewidth vs len(), I'll fix tomorrow

mathaou:
> 

I actually didn't do any kind of figuring based on individual string lengths. I left the original API intact. The widths of each column is defined by the parent application, so this only touches slice lengths.

bashbunni:
Thank you for the PR! It looks great. Would you mind outlining the what you changed for us to test against? I know it's all in the code, but it would be a huge time saver. 

I did a basic validation of the cell selection and navigation with `ctrl+t` - looks great, the only thing is the padding needs to get tweaked a bit (probably need to reset the styles when not in select cell mode) when we exit select cell mode back to the line selection. It seems to push the characters of the selected row off to the right by one character. Recording of behaviour attached.


https://user-images.githubusercontent.com/15822994/196605116-fe2ebd21-516c-4be0-a70b-e7c5ce14a851.mov

bashbunni (changes requested):
- [ ] fix text alignment when switching back from `CellSelect`

PR #266, opened 2022-10-06
feat(textarea): add Header field to display above the input

keevan:
Relates to charmbracelet/gum#80

So I foolishly realized that I've been playing around with the "View" in gum-land when I should have updated it here.

So I'll need to move some of that logic across I believe.


keevan:
Thanks for the feedback. I'll see what I can do.

If I'm understanding correctly, the parent model in this case would be `write.go` in the gum repo.

keevan:
Cool, so I think I wrapped my head around it. I'll close off this PR since it no longer needs to exist.

Thanks again for the push in the right direction.

PR #283, opened 2022-11-03
Expose TotalLineCount() and VisibleLineCount() methods within viewport

adaam2:
[As per discord discussion](https://discord.com/channels/1032368467246075914/1032368467988451425/1037767861030096976), exposes a `TotalLineCount()` and `VisibleLineCount()` methods in the viewport package.

### DIFF
--- a/viewport/viewport.go
+++ b/viewport/viewport.go
@@ -207,6 +207,16 @@ func (m *Model) LineUp(n int) (lines []string) {
 	return m.visibleLines()
 }
 
+// TotalLineCount returns the total number of lines (both hidden and visible) within the viewport.
+func (m Model) TotalLineCount() int {
+	return len(m.lines)
+}
+
+// VisibleLineCount returns the number of the visible lines within the viewport.
+func (m Model) VisibleLineCount() int {
+	return len(m.visibleLines())
+}
+
 // GotoTop sets the viewport to the top position.
 func (m *Model) GotoTop() (lines []string) {
 	if m.AtTop() {


meowgorithm:
Thanks, @adaam2. Per the discussion:

> Is there a particular reason why lines in viewport is not accessible outside of the package? I have found myself in a situation where it would be helpful to know how many lines there currently are in a viewport - hidden or visible:
>
> Don't think there's a method currently that allows you see the total number of lines - scroll percent is the nearest thing

 Just curious about your use case here: is there a particular reason you need the actual lines versus just the total number? Was assuming this PR would be something closer to something like this;

```go
func (m Model) TotalLineCount() int

func (m Model) VisibleLineCount() int
```

viewport/viewport.go
```diff
@@ -207,6 +207,11 @@ func (m *Model) LineUp(n int) (lines []string) {
 	return m.visibleLines()
 }
 
+// Lines returns all of the lines (both hidden and visible) within the viewport.
+func (m *Model) Lines() []string {
```
meowgorithm, line 211: @muesli: So we usually copy the type for getters. In the case of Bubble Tea, do you think it makes sense to instead start using pointer methods on Bubbles as an optimization?

adaam2:
@meowgorithm yeah I'd be happy with just the totals - figured the actual content of the lines may have been more useful for others going forward (maintaining snapshots of content etc, but then again could just capture that in other ways just as easily)

I'll change the PR unless you have had a change of heart.

viewport/viewport.go
```diff
@@ -207,6 +207,16 @@ func (m *Model) LineUp(n int) (lines []string) {
 	return m.visibleLines()
 }
 
+// TotalLineCount returns the total number of lines (both hidden and visible) within the viewport.
+func (m *Model) TotalLineCount() int {
```
meowgorithm, line 211: For now, let's remove the pointer receiver on this method.

meowgorithm:
@adaam2 Thank you! For now, let's just stick with the totals. Just a couple small requests, then we're good to go with this one.

adaam2:
@meowgorithm updated 👍🏻

meowgorithm:
Looks good. Thank you!

PR #287, opened 2022-11-07
Table: expose rows

marcantoineg:
## Description 📖 
Exposes a `table`'s rows through `func Rows() []Row`


maaslalani:
Thank you so much! This is looking great ❤️

PR #295, opened 2022-11-16
chore: apply appropriate GoDoc syntax to deprecated items

meowgorithm:
Just a little GoDoc housekeeping with regard to deprecate items.


meowgorithm:
Thanks, Muesli!!

PR #300, opened 2022-12-03
Fix Typo

yusufmalikul:


paginator/paginator.go
```diff
@@ -81,7 +81,7 @@ func (m *Model) SetTotalPages(items int) int {
 	return n
 }
 
-// ItemsOnPage is a helper function for returning the numer of items on the
+// ItemsOnPage is a helper function for returning the number of items on the
 // current page given the total numer of items passed as an argument.
```
muesli, line 85: The same typo is repeated in this line, as well as in `table/table.go`. Would be great to fix them all at once.

muesli (changes requested):

paginator/paginator.go
yusufmalikul: Right.
I fixed them too.

maaslalani (approved):

maaslalani:
Sorry, made a type in the commit message 🤦, will leave it as I don't want to force push to `master`

PR #302, opened 2022-12-06
feat(textarea) Add multiline placeholder

mikelorant:
Add the capability to show a multiline placeholder. Some refactoring was required to improve readability and improve logic.

End of line buffer character was only shown when line numbers were displayed which requires some verification whether this is the intended outcome. This change resolves this issue.

### DIFF
--- a/textarea/textarea_test.go
+++ b/textarea/textarea_test.go
@@ -326,6 +326,8 @@ func TestVerticalNavigationShouldRememberPositionWhileTraversing(t *testing.T) {
 }
 
 func TestView(t *testing.T) {
+	t.Parallel()
+
 	type want struct {
 		view      string
 		cursorRow int
@@ -1264,7 +1266,64 @@ func TestView(t *testing.T) {
 			},
 			want: want{
 				view: heredoc.Doc(`
-					>   1 .
+					>   1 H
+					>     e
+					>     l
+					>     l
+					>     o
+					>     ,
+				`),
+			},
+		},
+		{
+			name: "placeholder single line",
+			modelFunc: func(m Model) Model {
+				m.Placeholder = "placeholder the first line"
+				m.ShowLineNumbers = false
+
+				return m
+			},
+			want: want{
+				view: heredoc.Doc(`
+					> placeholder the first line
+					>
+					>
+					>
+					>
+					>
+					`),
+			},
+		},
+		{
+			name: "placeholder multiple lines",
+			modelFunc: func(m Model) Model {
+				m.Placeholder = "placeholder the first line\nplaceholder the second line\nplaceholder the third line"
+				m.ShowLineNumbers = false
+
+				return m
+			},
+			want: want{
+				view: heredoc.Doc(`
+					> placeholder the first line
+					> placeholder the second line
+					> placeholder the third line
+					>
+					>
+					>
+				`),
+			},
+		},
+		{
+			name: "placeholder single line with line numbers",
+			modelFunc: func(m Model) Model {
+				m.Placeholder = "placeholder the first line"
+				m.ShowLineNumbers = true
+
+				return m
+			},
+			want: want{
+				view: heredoc.Doc(`
+					>   1 placeholder the first line
 					>
 					>
 					>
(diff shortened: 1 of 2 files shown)


mikelorant:
@meowgorithm @maaslalani Can I get approval to run the workflow? Thanks.

mikelorant:
Can I get someone from @charmbracelet to review this. Thanks.

maaslalani:
Hey @mikelorant. Thank you so much for this PR.

I'm seeing a bug with how this PR is handling line numbers:

This is what the textarea should look like:
image


Here is what it looks like using this PR (rebased on main):

image

maaslalani:
The testing code I'm using can be found here: https://github.com/charmbracelet/bubbletea/tree/master/examples/textarea

mikelorant:
Just to clarify, line numbers only display if there is user input (or cursor) on that line?

mikelorant:
Just did some basic investigation and it seems the aim is that it should look like Vim with line numbers enabled. Which means only lines with content get a number.

maaslalani:
> Just did some basic investigation and it seems the aim is that it should look like Vim with line numbers enabled. Which means only lines with content get a number.

Yep, that's correct! Similar to vim!

mikelorant:
One more reference image with the cursor visible.

CleanShot 2024-01-31 at 18 51 23

Check the test tables and let me know if that is the output you'd expect. I had to make some executive decisions and tried to base everything I could on how it was displayed in Vim.

mikelorant:
@maaslalani Any chance you could do a review, if this is working correctly I'd like to get it merged otherwise if there are problems, lets me get started on fixing them.

maaslalani:
Hey @mikelorant, thanks again for your work on this one!

One issue I'm seeing is the CursorLine (if it has a background color, bleeds over to the cursor column)

image


Here's what it should look like:

image

mikelorant:
The default prompt which is being used here is defined at:
https://github.com/charmbracelet/bubbles/blob/master/textarea/textarea.go#L258C3-L258C59

Code as follows:
```
Prompt:               lipgloss.ThickBorder().Left + " ",
```

This has me thinking that both versions are wrong. The background colour should start at position 3 for each line.
```
|X123 Once upon a time...
|X    ~
|X    ~
```

The `X` space is part of that prompt and whatever styling that is applied to it is independent of everything else.

Thoughts?

mikelorant:
@maaslalani I'll quickly do the update to this pull request providing the original functionality, but I still believe both implementations have it wrong.

mikelorant:
Here are the two possibilities in one image:

CleanShot 2024-02-02 at 17 56 33

The top version is how it is currently implemented, however I believe the bottom version is how it should be. However a lowly minion here to service the wonder Charm team 😄 I have pushed up the version that conforms with how it is also displayed when text is entered.

mikelorant:
Heres the example showing the line style for the cursor line leaking out of the prompt "border".
CleanShot 2024-02-02 at 18 06 47

Prompt to use see this case:
```go
ti.Prompt = lipgloss.ThickBorder().Right + " "
```

maaslalani:
@mikelorant Thanks for the care you're putting into this PR! Yes, I see what you mean by the border. However, when the cursor line is defined with a background color I think it looks more correct (but maybe we will change this later since you are actually right about the correctness technically)

Also when a text area has a multiline placeholder I believe the cursor line background should be applied to the entire placeholder. In your screenshots it only applies to the first line.

mikelorant:
Here is a comparison of the two:

CleanShot 2024-02-03 at 08 58 06

I think it is very confusing to highlight more than the cursor line where there is a multiline placeholder. However, on the flip side it may help indicate that any typing will clear all that text away by keeping it "grouped".

How should I proceed?

maaslalani:
In my opinion the second option is correct here as the entire line is highlighted with the cursor line and if the user types that exact placeholder then the same highlighting will occur.

mikelorant:
@maaslalani Commit updated to highlight all lines of the placeholder.

mikelorant:
@maaslalani Can you please trigger the workflow. Thanks.

mikelorant:
@maaslalani Anything blocking this from merging? Like to get this one wrapped up.

meowgorithm:
Hey, @mikelorant! Maas is out traveling this week; I imagine he'll be able to take a look next week after he's back. He's the best person to give this one the a-ok.

mikelorant:
@meowgorithm Thanks for the update! Will put this on the backburner then.

mikelorant:
@maaslalani Any chance for review please?

maaslalani:
Hey @mikelorant, will try to get this in soon. Sorry for the delay!

mikelorant:
@maaslalani No problem, thanks for letting me know.

maaslalani:
Noticing a bug here with single line placeholders:

The end of line buffers shift when entering a character.

image


image

maaslalani:
Also if possible can we rebase this on `master` to ensure it works with the latest changes of `bubbles`.

mikelorant:
So I believe this is a bug with the existing textarea. Be aware I have only changed the code for the placeholder.

Here is what it looks like in Vim.
Vim

My unit tests for a placeholder is defined as the follows:
```go
{
	name: "single line with show line numbers",
	lines: []string{
		"the first line",
	},
	showLineNumbers: true,
	expected: []string{
		">   1 the first line",
		"> ~",
		"> ~",
		"> ~",
		"> ~",
		"> ~",
	},
},
```

As soon as you typed a letter you were no longer using the placeholder function and the normal textarea functionality applies.

Would we like to maintain the previous (flawed) implementation or correct it to be like Vim?

mikelorant:
> Also if possible can we rebase this on `master` to ensure it works with the latest changes of `bubbles`.

Rebased and confirmed tests still pass.

mikelorant:
Found the bug in `View()`.
```go
for i := 0; i < m.height; i++ {
	prompt := m.getPromptString(displayLine)
	prompt = m.style.Prompt.Render(prompt)
	s.WriteString(prompt)
	displayLine++

	if m.ShowLineNumbers {
		lineNumber := m.style.EndOfBuffer.Render((fmt.Sprintf(m.lineNumberFormat, string(m.EndOfBufferCharacter))))
		s.WriteString(lineNumber)
	}
	s.WriteRune('\n')
}
```

As you can see it always renders `LineNumberFormat` before `EndOfBufferCharacter`. Even though line number is not visible, it still adds the padding.

We can fix this via a separate pull request if you prefer.

mikelorant:
Fix is quite easy in `View()`.

We actually don't want line numbers on lines after the contents.

**Before**
```go
lineNumber := m.style.EndOfBuffer.Render((fmt.Sprintf(m.lineNumberFormat, string(m.EndOfBufferCharacter))))
```

**After**
```go
lineNumber := m.style.EndOfBuffer.Render(string(m.EndOfBufferCharacter))
```

I'll create 2 more pull requests.
1. Add unit tests for the textarea with content already entered.
2. Add a fix for the line numbers used with end of buffer character.

mikelorant:
**With placeholder**
Placeholder

**After text input**
Input

mikelorant:
Here is the commit I will use for the fix.
https://github.com/charmbracelet/bubbles/commit/3c707bfe5212af52903095408636a30180406c34

I want to get unit tests added first and then change the functionality, so pull request will come later.

mikelorant:
Can confirm that the unit test for the end of line buffer character is flawed.

```go
func TestRendersEndOfLineBuffer(t *testing.T) {
	textarea := newTextArea()
	textarea.ShowLineNumbers = true
	textarea.SetWidth(20)

	view := textarea.View()
	if !strings.Contains(view, "~") {
		t.Log(view)
		t.Fatal("Expected to see a tilde at the end of the line")
	}
}
```
https://github.com/charmbracelet/bubbles/blob/master/textarea/textarea_test.go#L419-L429

A simple `string.Contains()` isn't really testing anything useful.

How about we remove this unit test and replace it with some test tables that check a lot more and compare them to golden files?

maaslalani:
> Can confirm that the unit test for the end of line buffer character is flawed.
> 
> 
> 
> ```go
> 
> func TestRendersEndOfLineBuffer(t *testing.T) {
> 
> 	textarea := newTextArea()
> 
> 	textarea.ShowLineNumbers = true
> 
> 	textarea.SetWidth(20)
> 
> 
> 
> 	view := textarea.View()
> 
> 	if !strings.Contains(view, "~") {
> 
> 		t.Log(view)
> 
> 		t.Fatal("Expected to see a tilde at the end of the line")
> 
> 	}
> 
> }
> 
> ```
> 
> https://github.com/charmbracelet/bubbles/blob/master/textarea/textarea_test.go#L419-L429
> 
> 
> 
> A simple `string.Contains()` isn't really testing anything useful.
> 
> 
> 
> How about we remove this unit test and replace it with some test tables that check a lot more and compare them to golden files?

Instead of golden files I would prefer if we simply embed the whole string in the test.

maaslalani:
Thanks for the work on this one, this is looking great!

mikelorant:
@maaslalani Reworked the tests using the new test tables, cleans everything up nicely and reduces the amount of code added.

I'm hoping we have resolved all the issues now and the tests verify the actual outcomes we want.

maaslalani:
Thanks so much @mikelorant, reviewing now!

maaslalani:
If the placeholder is larger than the width we should wrap the placeholder:

image

mikelorant:
> That is, long placeholders should behave as multi-line placeholders.

Agreed, let me add a test case and fix for this one.

mikelorant:
@maaslalani This started to get complicated so I had to use `muesli/reflow` to handle wrap and wordwrap.

Have lots of test cases (include some edge cases) to verify this is working as expected.

mikelorant:
Added a few more test cases because I am paranoid of all the edge cases 😄

textarea/textarea_test.go
```diff
@@ -606,10 +608,370 @@ func TestView(t *testing.T) {
...
+				m.MaxWidth = 20
+				m.ShowLineNumbers = true
+
+				return m
+			},
+			expected: heredoc.Doc(`
+				>   1 12345678901234
+				>     5
+				>     12345678901234
+				>     5
```
maaslalani, line 962: I can't describe how awesome this test is! Thank you so much!

mikelorant: We need a lot more boundary checking in the tests.

Feel free to recommend any others you think you are useful (they are pretty quick to code).

mikelorant:
@maaslalani Don't believe we have any issues left with this pull request. Would love to get this merged as my own app depends on it.

meowgorithm:
We're waiting on https://github.com/muesli/reflow/pull/71 for this one, correct?

mikelorant:
@meowgorithm Not at all. This one is ready to go.

This has nothing to do with emojis or grapheme clusters.

mikelorant:
@maaslalani Rebase completed and all unit tests are passing.

🤞 that there are no more issues.

mikelorant:
@meowgorithm @maaslalani Can we please get this reviewed. Want to get this finalised so I can move off using my own fork.

maaslalani:
@mikelorant Will take a look today

maaslalani:
@mikelorant I'm still seeing the bug of long placeholders, let me know if I'm doing something incorrectly.

But, if I set the placeholder to:

```
	ti.Placeholder = "Once upon a time in a land far far away..."
```

Then the text is broken.

image

However if I add a new line it works as expected:

image


I wouldn't mind truncating the placeholder like we currently do if the user has not explicitly opted into the newline, if that easier.

textarea/textarea_test.go
```diff
@@ -672,10 +674,410 @@ func TestView(t *testing.T) {
...
+				m.Placeholder = "placeholder the first line that is longer than the max width"
+				m.MaxWidth = 40
+				m.ShowLineNumbers = false
+
+				return m
+			},
+			want: want{
+				view: heredoc.Doc(`
+					> placeholder the first line that is
+					> longer than the max width
```
maaslalani, line 845: I think we need to add a test where the `CursorLine` is styled like we see in the [bubbletea example](https://github.com/charmbracelet/bubbletea/blob/master/examples/textarea/main.go).

mikelorant: Lots more tests coming up.

mikelorant:
> I wouldn't mind truncating the placeholder like we currently do if the user has not explicitly opted into the newline, if that easier.

No truncating, will work out what is going on.

Just need lots more tests to find all these weird edge cases that keep popping up. Not sure why this feature has been so difficult to implement 😢 

Thanks for finding these problems.

mikelorant:
Found the bug and unfortunately it seems like we may have 2 bugs 😢 

Firstly, my bug is quite simple, I was using `m.MaxWidth` to determine the width instead of `m.width`. Easy fix.

However, this has uncovered something new with all the tests we have added.

Looking into the function `SetWidth`, there is a lot of work done to make sure the width removes the prompt and line numbers. However the last check caused an edge case.

```go
if m.MaxWidth > 0 {
    m.width = clamp(inputWidth, minWidth, m.MaxWidth)
} else {
    m.width = max(inputWidth, minWidth)
}
```

If the `MaxWidth` is smaller than `inputWidth` we pick `MaxWidth`. Cool? Cool. However, we never subtracted the prompt and line numbers from this. We need to pick the whether we use `inputWidth`, `minWidth` or `m.MaxWidth` first and then subtract the prompt and line numbers.

Hoping my logic here is correct, will keep hacking away and see where this leads.

mikelorant:
Going to add another problem, setting `MaxWidth` on the model does not call `SetWidth()` so updating of the model width is not being done.

I believe the order we need to follow is:
```go
ta := textarea.New()
ta.MaxWidth = 25
ta.SetWidth(20)
```

This might be a case where `MaxWidth` should have been a function so it then it could call `SetWidth()` before it returns.

Definitely need some guidance on how this should work.

mikelorant:
Sorted out all the problems, however this is now dependent on #496.

maaslalani:
Thanks for debugging the `SetWidth` and improving it in #496, I can take a look at this one when it is ready to go!

mikelorant:
Have fixed the conflicts and unit tests now pass.

Will run some manual tests to see if we have all the issues finally resolved 😄

mikelorant:
Seeing some new issues unfortunately, so not ready for review yet.

textarea-testing

mikelorant:
@maaslalani Actually, everything looks OK from both unit tests and manual tests. Hopefully you see the same results.

mikelorant:
@maaslalani Can we get another review of this one, I am hoping this one passes now that we have the `SetWidth` method fixed.

mikelorant:
@maaslalani Can you please take another look at this? I have rebased it from `master` and made the fixes required to resolve the conflicts.

maaslalani:
Thanks @mikelorant, I will take a look! Thanks for the reminder!

maaslalani:
Thanks @mikelorant, tested things out in the example and worked well. Thank you so much for the extensive test suite also, it really makes these PRs easier to merge.

Really appreciate the patience and great work ❤️

mikelorant:
Wahoo. Finally merged! Thank you so much for your efforts in making sure we keep the quality high.

PR #313, opened 2023-01-23
fix(textarea): app would crash if deleteWordRight was called at the end of line

infastin:


maaslalani:
Thank you so much, just tested this out and you're absolutely correct!

PR #314, opened 2023-01-24
fix(textarea): support pastes that end with a newline character

knz:
My previous fix did not properly handle a multi-rune input that *ends* with a newline character. In that case, the final newline was removed.

I found this using the test suite I built for textarea in bubbline at https://github.com/knz/bubbline/tree/main/editline/internal/textarea/testdata.

cc @maaslalani @muesli


maaslalani:
Thanks so much @knz! Very much appreciated!

PR #318, opened 2023-02-05
fix(textarea): fix data corruption after multi-line input

knz:
Found using the unit test suite at https://github.com/knz/bubbline

Previously, when inserting multiple lines of input at once, the underlying Go data array was incorrectly shared in memory across multiple rows of the textarea. This was causing input corruption, where a user trying to modify one line would see their text added on the next line as well.

This patch fixes it.

cc @maaslalani


maaslalani:
Thank you so much @knz, this is really appreciated ❤️

maaslalani (approved):

PR #201, opened 2022-08-05
Feat(list): enable filtering programmatically

fkhadra:
Hey! This PR aims to address #85. The names for the method I've added are coming from #129. 

- set filter value programmatically
- enable filtering programmatically. Apply filter if needed

https://user-images.githubusercontent.com/5574267/183026053-8169a28c-b862-467b-bdf6-d538769148df.mov

Below is the code used in the video. I tried to keep it as simple as possible
```go
package main

import (
	"github.com/charmbracelet/bubbles/list"
	tea "github.com/charmbracelet/bubbletea"
)

type item struct {
	title string
}

func (i item) Title() string       { return i.title }
func (i item) Description() string { return i.title }
func (i item) FilterValue() string { return i.title }

type Model struct {
	list list.Model
}

func New() Model {
	items := []list.Item{
		item{title: "red"},
		item{title: "blue"},
		item{title: "orange"},
		item{title: "purple"},
		item{title: "green"},
		item{title: "magenta"},
	}
	l := list.New(items, list.NewDefaultDelegate(), 100, 100)

	// uncomment to set default filter value
	// l.SetFilterValue("purple")

	return Model{
		list: l,
	}
}

func (m Model) Init() tea.Cmd {
	// used to enable filtering programmatically
	return list.EnableLiveFiltering
}

func (m Model) Update(msg tea.Msg) (tea.Model, tea.Cmd) {
	var (
		cmd  tea.Cmd
		cmds []tea.Cmd
	)

	switch msg := msg.(type) {
	case tea.WindowSizeMsg:
		m.list.SetSize(msg.Width, msg.Height)
	}

	m.list, cmd = m.list.Update(msg)
	cmds = append(cmds, cmd)

	return m, tea.Batch(cmds...)
}

func (m Model) View() string {
	return m.list.View()
}

```

I did not implement a way to disable the filter programmatically as I believe this can already be done using `ResetFilter`.

I've also tested my implementation with a custom input by setting `l.SetShowFilter(false)`.

Issue #85, opened 2021-12-21
Activate filtering programmatically


revett:
## Problem

I'm looking to enable filtering for my `github.com/charmbracelet/bubbles/list` component by default so that as soon as the list is rendered I can start to type, without needing to press the filter shortcut (`/`) first. 

## Recreate

- Use the `list-default` example within the [charmbracelet/bubbletea](https://github.com/charmbracelet/bubbletea/blob/master/examples/list-default/main.go) repo
- Set `m.list.SetFilteringEnabled(true)` on line 82
- Note that this does not do anything

I'm relatively new to the package so I think this needs to be done at a different point within the model lifecycle, rather than at the point of creation. Any help would be great 👍

meowgorithm:
Hi! [SetFilteringEnabled](https://pkg.go.dev/github.com/charmbracelet/bubbles@v0.9.0/list#Model.SetFilteringEnabled) is actually for enabling and disabling filter functionality, not quite for triggering filtering. Based on my quick skimming of the code _think_ we’ll need to add a method for you to be able to activate the filter UI programmatically.

Anyway, it’s a good idea, and we should definitely add it (surprised we haven’t, actually), though it will probably happen after the new year.

joaom00:
Any update around this?

akgunberk:
I am facing the same issue. I just want to view filter input with a different key msg and use it for another purpose rather than filtering. But now way to achieve this. m.list.SetShowFilter() doesn't work and does not replace title. Any news on this issue?

usrme:
Same thoughts as @akgunberk: I would ideally just want to co-opt the same screen space the filtering currently takes up, but activate it with a different key and then just capture the input without any filtering.

bashbunni:
Hey @revett I believe this is fixed in https://github.com/charmbracelet/bubbles/pull/335 - could you please let us know if this is still an issue? It sounds to me like it's the same feature request

PR #335, opened 2023-02-21
Add SetFilterText and SetFilterState

taigrr:
Currently, there's no way to allow a developer to prefill a list filter reliably. A workaround of using Program.Send works, but is limiting.

This PR introduces two new functions, SetFilterText and SetFilterState which fix this issue and allow for finer control over the list's initial filter state.

Willing to make whatever style changes etc. are required.

### DIFF
--- a/list/list.go
+++ b/list/list.go
@@ -10,16 +10,17 @@ import (
 	"strings"
 	"time"
 
-	"github.com/charmbracelet/bubbles/help"
-	"github.com/charmbracelet/bubbles/key"
-	"github.com/charmbracelet/bubbles/paginator"
-	"github.com/charmbracelet/bubbles/spinner"
-	"github.com/charmbracelet/bubbles/textinput"
 	tea "github.com/charmbracelet/bubbletea"
 	"github.com/charmbracelet/lipgloss"
 	"github.com/muesli/reflow/ansi"
 	"github.com/muesli/reflow/truncate"
 	"github.com/sahilm/fuzzy"
+
+	"github.com/charmbracelet/bubbles/help"
+	"github.com/charmbracelet/bubbles/key"
+	"github.com/charmbracelet/bubbles/paginator"
+	"github.com/charmbracelet/bubbles/spinner"
+	"github.com/charmbracelet/bubbles/textinput"
 )
 
 // Item is an item that appears in the list.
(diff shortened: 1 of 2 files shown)


taigrr:
This also addresses #85 but in a different way than #201

taigrr:
thanks for the tests (both manual and programmatic) @k-x7 !

list/list.go
```diff
@@ -255,12 +256,40 @@ func (m *Model) SetShowTitle(v bool) {
 	m.updatePagination()
 }
 
+// SetFilterText explicitly sets the filter text without relying on user input.
+// It also sets the filterState to a sane default of FilterApplied, but this
+// can be changed with SetFilterState
+func (m *Model) SetFilterText(filter string) {
+	m.filterState = Filtering
+	m.FilterInput.SetValue(filter)
+	cmd := filterItems(*m)
```
andrewebdev, line 265: Should this maybe be returned by the functions? If a command is sent by a function, surely it might be prudent to send that along?

andrewebdev: Basically `SetFilterText()` should return a `tea.Cmd`?

taigrr: No, we call it and handle it. The model might not have been initialized yet. We can't assume the program is already running.

andrewebdev: ah ok that makes sense.

taigrr:
@caarlos0 / @maaslalani  any chance I could get this revisited?

lorenzo-milicia:
This implementation does not set the `filteredItems` list to be the whole list when the state is set to `Filtering`, and the `FilterInput` has empty string as value. I would add the same code found in the switch to the `Filtering` state by key press:
```
if m.FilterInput.Value() == "" {
// Populate filter with all items only if the filter is empty.
m.filteredItems = m.itemsAsFilterItems()
}
```

ajayd-san:
Hey, 
Is there a reason why this is not yet merged? I'd like to work on it if no one else is.

caarlos0 (approved):

meowgorithm (approved):

taigrr:
Nearing the 1 year anniversary of my last bump request :)

Does the second approval mean we're closer to a merge?

caarlos0:
thanks @taigrr , sorry for the delay marging this

taigrr:
Awesome. No worries, I know y'all are busy and this adds exported functionality to the API, it should be vetted thoroughly :)

meowgorithm:
Thanks for your patience with this one @taigrr ❤️

PR #339, opened 2023-02-26
refactor(table): remove unused cmds variable

stefanbildl:
the `cmds` variable in table's Update method is unused. 
I replaced it with `nil`.


maaslalani:
Thank you so much!

Issue #246, opened 2022-09-17
Feature request: Style table rows / cells individually


SkyLeite:
I would like to be able to provide styling for rows and / or cells individually. Currently the only way to do this is through the `Selected` property, but that approach is limited in the sense that you can only have two styles (one for the base, and one for the selected row).

maaslalani:
Hey, this is a good idea and natural extension of the table. I think we should probably introduce a row that can take any struct that implements the `Stringer` interface so that you can pass lipgloss Styles and anything else for displaying the table.

pja237:
Hey all,
@maaslalani this is my attempt at this feature, goal for me was:
1. be backward compatible, all the code using the table package continues to work as-is
2. not have to change too much of the table.go code (which tbh. i don't know how to achieve by changing the Row to an interface, i end up having to refactor half of the package, or break no 1. and do ton of code changes in the existing app 😆 )

and this is the best i could come up with, if you'd be interested to take a look and let me know what you think of this approach (it's a local, older copy of tables code i'm still using, but i can easily PR this here)
1. i've composed a new `CustomStyles` struct in the table model
https://github.com/CLIP-HPC/SlurmCommander/blob/alt_table_2/internal/table/table.go#L22

which holds interfaces for Cell/Row/Selected functions.
```
// int is the row/col number, row is the row content, used in the methods to generate custom styles
type CustomStyle interface {
	Style(int, Row) lipgloss.Style
}
```
https://github.com/CLIP-HPC/SlurmCommander/blob/alt_table_2/internal/table/table.go#L28-L32

2. modified the renderRow() method to include checks if functions are present in the model and do the "overlaying" of cell<-row<-selected styles (using `.Inherit()`) so we can compose them.
https://github.com/CLIP-HPC/SlurmCommander/blob/alt_table_2/internal/table/table.go#L406-L454

This way, separate styles for a single Cell/Column of cells/Row/Selected rows can be programmatically done, like i'm doing here (first working prototype):
https://github.com/CLIP-HPC/SlurmCommander/blob/alt_table_2/internal/model/tabs/clustertab/clustertabtable.go#L69-L90
<img width="964" alt="Screenshot 2023-02-24 at 13 54 42" src="https://user-images.githubusercontent.com/11192275/221183975-cd9a3616-642c-4d48-a8e2-64183244570b.png">

jlec:
@pja237  Could you create a PR so that we can work together on refining your idea?

pja237:
[...]

PR #340, opened 2023-02-27
feat(table): More granular table styles #246

pja237:
@jlec et al.
PR regarding #246 

This was an idea of how one might extend the table styling without causing too much havoc in the table code and to maintain backwards compatibility.
I'm more then happy to receive feedback, ideas and pointers on how to make this better and eventually merged.

### DIFF
--- a/table/table.go
+++ b/table/table.go
@@ -19,10 +19,21 @@ type Model struct {
 	cursor int
 	focus  bool
 	styles Styles
+	CustomStyles
 
 	viewport viewport.Model
 	start    int
 	end      int
+}
+
+type CustomStyles struct {
+	CustomCellStyle     CustomStyle
+	CustomRowStyle      CustomStyle
+	CustomSelectedStyle CustomStyle
+}
+
+type CustomStyle interface {
+	Style(int, Row) lipgloss.Style
 }
 
 // Row represents one line in the table.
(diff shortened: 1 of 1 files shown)


jlec:
This week is nuts. Hopefully, on the weekend, I will be able to tinker with your code

jlec:
Works for me. docs are missing.

bashbunni:
Part of me wonders if it would be more streamlined to extend the existing `Styles` struct to include an optional value for `Row`. It looks like there are already options for `Selected` and `Cell` in the `Styles` struct. The user would be able to set these styles to whatever `lipgloss.Style` definitions they'd like when they initialize the table. 

The `Row` field could probably just initialize a new style (`lipgloss.NewStyle()`), so all fields could call lipgloss.Render without issue and users can add their own customizations without it breaking. Let me know your thoughts on that

bashbunni (changes requested):
Thank you so much for the PR! Looking forward to seeing the updates. It'll be great to be able to customize these tables further

maaslalani:
Hey @pja237, thanks so much for this contribution.

Our plan is to replace the table renderer with Lip Gloss tables which will allow fully customizable Cell and Row styles with the `StyleFunc`.

I am going to close this PR for now (doing some cleaning up) but feel free to re-open this or create new issues / PRs if needed!

maaslalani:
Here is the Lip Gloss table documentation for reference:

https://github.com/charmbracelet/lipgloss/tree/master?tab=readme-ov-file#rendering-tables

PR #351, opened 2023-03-11
fix: add bounds checking to the selectedrow func

MikaelFangel:
### Changes:
 * This adds bounds checking to the function returning the row under the cursor. By adding the check, it avoids that the program crashes when calling the function and the cursor is out of bounds.

### Still an issue:
 * The cursor can still become out of bounds values, which is a problem in itself.

### DIFF
--- a/table/table.go
+++ b/table/table.go
@@ -266,10 +266,10 @@ func (m *Model) UpdateViewport() {
 // SelectedRow returns the selected row.
 // You can cast it to your own implementation.
 func (m Model) SelectedRow() Row {
-	// Check if the cursor is within bounds else return  nil
-	if !(m.cursor >= 0 && m.cursor < len(m.rows)) {
+	if m.cursor < 0 || m.cursor >= len(m.rows) {
 		return nil
 	}
+
 	return m.rows[m.cursor]
 }
 
--- a/table/table.go
+++ b/table/table.go
@@ -267,10 +267,10 @@ func (m *Model) UpdateViewport() {
 // You can cast it to your own implementation.
 func (m Model) SelectedRow() Row {
 	// Check if the cursor is within bounds else return  nil
-	if m.cursor < len(m.rows) && m.cursor >= 0 {
-		return m.rows[m.cursor]
+	if !(m.cursor >= 0 && m.cursor < len(m.rows)) {
+		return nil
 	}
-	return nil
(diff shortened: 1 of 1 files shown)


muesli:
Looks reasonable. Also noticed that `m.cursor` can become negative during debugging.

muesli:
@MikaelFangel Note the linter warning.

MikaelFangel:
@muesli yep, just noticed and have fixes it now.

table/table.go
```diff
@@ -266,7 +266,11 @@ func (m *Model) UpdateViewport() {
 // SelectedRow returns the selected row.
 // You can cast it to your own implementation.
 func (m Model) SelectedRow() Row {
-	return m.rows[m.cursor]
+	// Check if the cursor is within bounds else return  nil
+	if m.cursor < len(m.rows) && m.cursor >= 0 {
```
muesli, line 270: May I suggest swapping the logic around? We should check for `m.cursor` being out of bounds as the exception and keep the "happy path" on the left edge.

MikaelFangel: You mean? 
```suggestion
	if m.cursor >= 0 && m.cursor < len(m.rows) {
```

table/table.go
```diff
@@ -266,7 +266,11 @@ func (m *Model) UpdateViewport() {
...
+		return m.rows[m.cursor]
+	}
+	return nil
```
MikaelFangel, line 273: Like this?
```suggestion
	if !(m.cursor >= 0 && m.cursor < len(m.rows)) {
		return nil
	}
	return m.rows[m.cursor]
```

table/table.go
MikaelFangel: Great suggestion.

table/table.go
muesli: I'd probably go with:

```go
if m.cursor < 0 || m.cursor >= len(m.rows) {
    return nil
}

return m.rows[m.cursor]
```

Should be a little more digestible for future readers.

MikaelFangel: I see I've added a commit with the suggested change. Thank you for the comments. (:

MikaelFangel: And sorry for the mess with all the commits 😅

muesli:
Thank you, @MikaelFangel!

PR #369, opened 2023-04-14
feat(table): Add column return function

abeleinin:
Same as Rows function but for Columns.


prgres (approved):

maaslalani:
Thank you so much for the contribution @abeleinin! ❤️

PR #370, opened 2023-04-22
feat(textarea): make max width/height configurable

knz:
Over at CockroachDB we're finding needs for textareas with hundreds of input lines.

cc @maaslalani

### DIFF
--- a/textarea/textarea.go
+++ b/textarea/textarea.go
@@ -21,8 +21,8 @@ const (
 	defaultHeight    = 6
 	defaultWidth     = 40
 	defaultCharLimit = 400
-	maxHeight        = 99
-	maxWidth         = 500
+	defaultMaxHeight = 99
+	defaultMaxWidth  = 500
 )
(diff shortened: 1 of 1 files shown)


mmitchellg5:
👍

maaslalani:
Thanks @knz, out of curiosity are you noticing significant performance degradations with hundreds of input lines?

knz:
Lol yes, it's pretty bad.

But the alternative is even worse: with a fixed limit, any content entered on the last line misbehaves:

- newlines are ignored, so if a multi line content is pasted the newlines appear to be deleted and the content gets concatenated on a single line.

- then, if the content runs into the width limit as well, all remaining characters get outright ignored

Ideally we could change the UX to have a prompt "are you sure" when someone pastes a lot of data. But i imagine it would be a lot of work. Removing the limit and just let things be slow is much simpler to implement and also users learn the consequences quickly.

PR #373, opened 2023-04-28
Bugfix/render off by one

squrki:
Bug: The top folder re-renders when going into a subfolder. It happens when the subfolder contains a larger number of files than can be rendered on one screen.

file-picker

Fix:
file-picker-fixed

cc: @maaslalani @bashbunni


maaslalani:
Ah nice find, thank you so much!

PR #374, opened 2023-05-01
fix: filepicker on empty directory

caarlos0:
refs https://github.com/charmbracelet/gum/issues/343


maaslalani:
Thanks, @caarlos0! Testing now!

PR #380, opened 2023-05-30
Dual pane model

squrki:
Adds dualPane functionality as discussed. Please see demo below:

dualPane

cc: @maaslalani @taigrr @bashbunni


taigrr:
Hmm, looks like the linter is acting up, it's complaining about a line of code that's no longer there ¯\\\_(ツ)_/¯

taigrr:
@squrki can you rerun your tests with the merged-in funcs added by @meowgorithm ?

squrki:
> @squrki can you rerun your tests with the merged-in funcs added by @meowgorithm ?

I confirm that it works the same as before.

bashbunni:
Hey, thank you so much for this, and sorry for the delay! We're leaning towards not merging this due to the maintenance cost. It also adds rigidity in a case where it would likely be sufficient to provide guidance, yet allow the users to decide on the implementation themselves. We're a small team, so we have to be very careful about what abstractions we want to officially support.

If you wanted to make the bubble available to others, you can put it in your own repo and we would be happy to feature it in the [additional bubbles section](https://github.com/charmbracelet/bubbles?tab=readme-ov-file#additional-bubbles). Just create a PR with a link to your repo in that section of the README :)

PR #386, opened 2023-06-13
Problem: can't determine if the progress bar stopped updating

yrashk:
This would be useful to know to stop displaying it after the user was shown the progress bar to render to its end.

Solution: expose `IsComplete` method is based on what's been drawn

### DIFF
--- a/progress/progress.go
+++ b/progress/progress.go
@@ -211,8 +211,7 @@ func (m Model) Update(msg tea.Msg) (tea.Model, tea.Cmd) {
 		}
 
 		// If we've more or less reached equilibrium, stop updating.
-		dist := math.Abs(m.percentShown - m.targetPercent)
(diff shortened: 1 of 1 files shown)


progress/progress.go
```diff
@@ -362,6 +361,12 @@ func max(a, b int) int {
 	return b
 }
 
+// IsComplete returns true if the progress bar reached equilibrium
```
meowgorithm, line 364: Please also end the comment with a period here.

meowgorithm:
Look good. Thank you!

yrashk:
Thanks! What's your release-cutting cadence/policy?

meowgorithm:
We probably won’t cut a new release for a bit, but in the meantime you can pin to `master`, which is effectively the same as if we cut a release today.

PR #394, opened 2023-07-18
(bug): Fix slice calculation when inserting string

hopefulTex:
Fixed calculation error made when using `InsertString(string)` or true-pasting.

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -1,4 +1,4 @@
-module github.com/hopefulTex/sticky-placeholder
+module github.com/charmbracelet/bubbles
 
 go 1.17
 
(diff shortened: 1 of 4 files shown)


meowgorithm:
Thank you for the PR! Would you mind also updating this in `textinput` as well?

hopefulTex:
> Thank you for the PR! Would you mind also updating this in `textinput` as well?

Done and done!

hopefulTex:
Last 6 commits were adding a new feature, realizing it got added to the wrong branch, and subsequently removing said feature.

PR #396, opened 2023-07-22
perf(list): reduce allocations in `filterItems`

naglis:
As the number of items in `targets` is known, use `make` with the expected size instead of appending, which requires more allocations as the `targets` slice grows.


bashbunni (approved):
Looks good, thanks for the PR

PR #400, opened 2023-08-01
feat(table): expose `start` and `end` fields via getter methods

pyrooka:
This PR exposes the `start` and the `end` fields from the table model struct.

Not having setters is fine because that would break the inner logic, but handling
mouse clicks on rows is a bit inconvenient because the mouse events contain the 
relative position but to select the actual row under the (mouse)cursor, the absolute
row index is needed. With exporting these - mainly the start - it should be as easy
as `table.Start() + mouseMsg.Y`.

### DIFF
--- a/table/table.go
+++ b/table/table.go
@@ -321,6 +321,16 @@ func (m *Model) SetCursor(n int) {
 	m.UpdateViewport()
 }
 
+// Start returns the index of the first rendered row.
+func (m *Model) Start() int {
+	return m.start
+}
+
+// End returns the index of the last rendered row.
+func (m *Model) End() int {
+	return m.end
+}
+
 // MoveUp moves the selection up by any number of rows.
 // It can not go above the first row.
 func (m *Model) MoveUp(n int) {


pyrooka:
Can someone review this? I don't want to bother maintainers by tagging. :)

meowgorithm:
Hi, there. Just curious, what your use case for this?

pyrooka:
I'm working on a TUI app which has to support mouse clicks. Currently - when the content is scrolled - there is no way to get the offset of the first rendered row. Only the selected row is available (under the cursor), but that's not what I'd like to get.
I've made a small hack which works, but it's definitely not ideal :)
```go
rowOffset := int(reflect.ValueOf(&m.tableData.table).Elem().FieldByName("start").Int())
```

maaslalani:
Hey @pyrooka, thank you for the PR. We're working on swapping out the implementation of the Bubble Table with using Lip Gloss tables:

That being said, we'll want to make sure the API is as small as possible so we limit breaking changes as much as possible. I will close this PR for now but we can revisit after we introduce Lip Gloss tables as the rendering for the bubble.

PR #403, opened 2023-08-03
feat: add option to set a custom lipgloss renderer

aymanbagabas:
Lip Gloss defaults to `os.Stdout` to detect the background color and color profile of the terminal. This can be a problem when the application is using a different output stream like `os.Stderr` or a remote session like in a SSH server.

When using an SSH server, usually the server provides to the session a `$SSH_TTY` environment variable pointing to the allocated PTY to be used for the application. The value of `$SSH_TTY` points to the terminal TTY
  FD that _should_ be used to run the application.

Starting with Lip Gloss v0.7, you can use custom renderers to specify a different Lip Gloss output to detect the background color and color profile. This PR adds the necessary "glue" to the Bubbles in order to change the renderer being used to detect the color profile and background.


caarlos0 (approved):

textarea/textarea.go
```diff
@@ -111,14 +111,14 @@ type LineInfo struct {
...
+// Styles that will be applied to the text area.
 //
-// Style can be applied to focused and unfocused states to change the styles
+// Styles can be applied to focused and unfocused states to change the styles
 // depending on the focus state.
 //
 // For an introduction to styling with Lip Gloss see:
 // https://github.com/charmbracelet/lipgloss
-type Style struct {
+type Styles struct {
```
meowgorithm, line 121: Any particular reason for the change here other than aesthetics?

aymanbagabas: Consistency with other bubbles. The type alias should take care of backward compatibility.

maaslalani:
@aymanbagabas I think we can close this one?

caarlos0:
looks like yes @aymanbagabas , this is gonna be fixed by bubbletea v2, right?

aymanbagabas:
> looks like yes @aymanbagabas , this is gonna be fixed by bubbletea v2, right?

Right, with the current v2, we'd still need to pass the `ctx` down to the bubbles to create the styles :/

caarlos0:
yeah, so imo lets close this

PR #407, opened 2023-08-14
Support suggestions and autocompletion in textinput

toadle:
Continuation of discussion from here: https://github.com/charmbracelet/bubbles/pull/140

### DIFF
--- a/textinput/textinput.go
+++ b/textinput/textinput.go
@@ -265,7 +265,7 @@ func (m *Model) SetSuggestions(suggestions []string) {
 		m.suggestions = append(m.suggestions, []rune(s))
 	}
 
-	m.refreshMatchingSuggestions()
+	m.updateSuggestions()
 }
 
 // rsan initializes or retrieves the rune sanitizer.
@@ -626,7 +626,7 @@ func (m Model) Update(msg tea.Msg) (Model, tea.Cmd) {
 
 		// Check again if can be completed
 		// because value might be something that does not match the completion prefix
-		m.refreshMatchingSuggestions()
+		m.updateSuggestions()
 
 	case pasteMsg:
 		m.insertRunesFromUserInput([]rune(msg))
@@ -818,8 +818,8 @@ func (m *Model) canAcceptSuggestion() bool {
 	return len(m.matchedSuggestions) > 0
 }
 
-// refreshMatchingSuggestions refreshes the list of matching suggestions.
-func (m *Model) refreshMatchingSuggestions() {
+// updateSuggestions refreshes the list of matching suggestions.
+func (m *Model) updateSuggestions() {
 	if !m.ShowSuggestions {
 		return
 	}
--- a/textinput/textinput.go
+++ b/textinput/textinput.go
@@ -40,8 +40,8 @@ const (
 type ValidateFunc func(string) error
 
 // Triggered when a completion is requested.
-type completionMsg struct {
-	completion string
+type suggestionsMsg struct {
+	suggestions []string
 }
 
 // KeyMap is the key bindings for different actions within the textinput.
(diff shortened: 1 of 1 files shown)


maaslalani:
This is awesome @toadle! Thank you so much!

toadle:
@maaslalani You think we can get this merged?

There was one thing that I was still thinking about: If there are several suggestions populated

- "Hello World"
- "Hello You"
- "Foo bar"

and you currently have "Hello" typed in, you can switch through all three. Ending up in the last situation where there is nothing to complete when having "Foo bar" selected, because "Hello" can not be completed to "Foo bar".

Should the autocomplete work in a way where the input also looks for the next possible suggestion if the current one can not be used to complete? Effectively switching the current selected suggestion.

The otherwise to do it would be to always switch the possible suggestions externally.

I kind of have the feeling that it should cause it'll be more useful that way. But I also would like to get this merged ;-)

toadle:
Anyway, forget my last comment. I just fixed it here: https://github.com/charmbracelet/bubbles/pull/407/commits/2d16f32176f821a080f9ca7584c0798c476f033c

Should be good to go. I hope we can merge this, so that it doesn't get stale again :-)

toadle:
Little preview: 

CleanShot 2023-08-16 at 00 02 06

meowgorithm:
@toadle so what you're saying is that you're building a bubble tea integration called `toadle-copilot`, right?

toadle:
@meowgorithm Unsure if pun or not :-)?!

meowgorithm:
@toadle Totally joking around. This looks awesome!

toadle:
@maaslalani What do you say?

maaslalani:
Going to try and get this merged today! Thanks so much for all your work @toadle.

toadle:
@maaslalani I saw that you removed the Msg to set suggestions in the recent commit. Please remember that suggestions most likely will get stemmet asynchronously by the result of another cmd. We discussed that over on the other PR. Would you still set them via message then?

maaslalani:
> @maaslalani I saw that you removed the Msg to set suggestions in the recent commit. Please remember that suggestions most likely will get stemmet asynchronously by the result of another cmd. We discussed that over on the other PR. Would you still set them via a func-call then?

Yep, I will add the asynchronous setting back since many people may get their autocompletion data from an API or external source. But we should definitely also allow setting the suggestions statically since a lot of people may already have the values beforehand when creating the textinput bubble.

maaslalani:
Hey @toadle looks like this PR will need some adjustments to get double width runes working: I can handle that soon!

image

toadle:
@maaslalani OK, that double-width rune thing is weird. I tried to fix that myself just now but always end up with the same problem. I'll also have a look if I can fix that. 

I pushed a little other change where I expose a function that can help customize to complete behaviour (e.g. if you want to complete just up to the next "space" or "-"). I think it'd be good to have that - I can think of several use-cases.

toadle:
@maaslalani Finally understood how `[]rune` vs. `string` works in Go. Refactored everything to work with `[]rune` - much better now. Should work fine with "charm热爱开源" and sorts now.

maaslalani:
Amazing @toadle! I will take a look and merge this soon 🎉
Thanks again for all your work!

maaslalani:
@toadle after doing a bit of testing it's definitely possible to use the SetSuggestions stuff with custom commands on the developer end. So I say we merge this as is.

For example, a user can do the following:

```go
type setSuggestionsMsg []string

func loadSuggestions() tea.Msg {
	time.Sleep(5 * time.Second) // Fetch from a data source.
	return setSuggestionsMsg(pokemon[0])
}

// ... Update method.
	case setSuggestionsMsg:
		m.textInput.SetSuggestions(msg)

```

Since the user can implement a basic `Cmd` for asynchronous data / completion fetching I think this is enough to enable this functionality for now. If necessary / requested we can add a first class API for async completion but I think the current way is more customizable for the developer.

maaslalani:
Also tested with double width runes and everything is working great!

PR #412, opened 2023-08-25
feat: Allow to mark colums as hidden. This might be helpful to hide s…

deicon:
Tables might contain more than needs to be visible to the user. 
Consider a table like 

ID  PrimaryKey          UserName
1    12392839281.    My_Username

Once displayed to the user, we want the user to select a row but don't want to expose the PrimaryKey. This is only used after
user selection for further processing. 

This can then be used to enable https://github.com/charmbracelet/gum/issues/320

### DIFF
--- a/table/table.go
+++ b/table/table.go
@@ -30,8 +30,9 @@ type Row []string
 
 // Column defines the table structure.
 type Column struct {
-	Title string
-	Width int
+	Title  string
+	Width  int
+	Hidden bool
 }
 
 // KeyMap defines keybindings. It satisfies to the help.KeyMap interface, which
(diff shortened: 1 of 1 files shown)


deicon:
@maaslalani any chance this gets a review any time soon ?

maaslalani:
> @maaslalani any chance this gets a review any time soon ? 
> 
> 

Hey @deicon, because of the new Lip Gloss tables (https://github.com/charmbracelet/lipgloss/releases/tag/v0.9.0) we're currently integrating that renderer into Bubbles which will then be used in Gum. So we should review / revisit this feature after the new table renderer is introduced in Bubbles and Gum.

maaslalani:
Hey @deicon, we recently introduced hiding a column by setting the width to 0:

* https://github.com/charmbracelet/bubbles/pull/465

Let us know if this solves your use case, feel free to re-open this PR or create a new issue if it doesn't!

deicon:
@maaslalani Kind of. Mostly I was trying to fix https://github.com/charmbracelet/gum/issues/320
by a PR in GUM. to make this even better, I would love to hide some columns but still use values of columns as return values from the gum table command. https://github.com/charmbracelet/gum/pull/415

PR #413, opened 2023-08-30
refactor(textinput): reduce allocations

naglis:
As the number of items in `suggestions` and `m.suggestions` is known, we can use `make` with the expected size instead of appending, which requires more allocations as the slice grows.


bashbunni (approved):

maaslalani:
Thanks so much for the PR @naglis, very much appreciated ❤️

PR #418, opened 2023-09-10
feat(textarea): add help to textarea key bindings

TravisYeah:
Adding help text to textarea key bindings


maaslalani:
Thanks so much for the contribution @TravisYeah!

PR #419, opened 2023-09-18
feat(viewport): add SetLines method in Viewport

semihbkgr:
In certain use-cases I encountered, setting content by lines proved to be more advantageous both in terms of functionality and performance.

### DIFF
--- a/viewport/viewport.go
+++ b/viewport/viewport.go
@@ -99,18 +99,18 @@ func (m Model) ScrollPercent() float64 {
 }
 
 // SetContent sets the content by normalizing line endings and splitting the content by lines.
-// For high performance rendering, the Sync command should also be called.
-func (m *Model) SetContent(s string) {
-	// Normalize line endings to '\n'
-	s = strings.ReplaceAll(s, "\r\n", "\n")
-	// Split the string into lines and set them in the Model
-	m.SetLines(strings.Split(s, "\n"))
-}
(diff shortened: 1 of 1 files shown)


meowgorithm:
This is fairly off the cuff, but I wonder if `SetContent` would be better off accepting variadic args rather than increasing surface area of the API.

```go
func (m *Model) SetContent(s string) // before
func (m *Model) SetContent(line ...string) // after
```

semihbkgr:
@meowgorithm do you mean that the API should accept only lines instead of whole content?

I think it will break the backward-compatibility for anyone using the Viewport model and it may not be suitable for every use case.

meowgorithm:
I actually mean that `SetContent` would be able to accept multiple args, joining them with newlines as a convenience. This would be compatible with the current API.

```go
vp := viewport.New(80, 40)
viewport.SetContent("a\nb", "c")
fmt.Println(viewport.View()) // outputs "a\nb\nc"
```

caarlos0:
I agree with @meowgorithm, probably having only one `SetContent` is best...

we can then iterate over each element, replace `\r\n` with `\n`, split by `\n` and add it as we currently do.

semihbkgr:
Great, then we’ve agreed on a single function, which will accept variadic arguments.

caarlos0:
this is fixed in the `v2-exp` branch (added a `SetContentLines` method).

Thanks for the PR :)

Issue #420, opened 2023-09-27
placeholder view Unicode character encoding error


lc-1010:
**Describe the bug**
 `Placeholder` View renders got Unicode character encoding error

**Setup**
use `textinput.Model`set   Unicode character is the Placeholder 
```
inputs= textinput.New()
 inputs.Placeholder = "馹ABC"
...
m.input.View()
...
 ���ABC
```

lc-1010:
You're welcome! I'm glad to hear that you were able to reproduce the issue. 

Whenever you have a moment ,could you kindly review and consider merging my [PR](https://github.com/charmbracelet/bubbles/pull/421) ? Thank you for your time.
 @meowgorithm

meowgorithm:
Oh brilliant, I totally missed that you have have a patch. Let's get this done!

meowgorithm:
Fixed by #421 🙌

PR #421, opened 2023-09-27
fix: Unicode character encoding error

lc-1010:
Unicode character encoding error
#420


meowgorithm:
Looks good—thank you!

Closes #420 🌿

PR #425, opened 2023-10-27
feat(textinput): Autocomplete when value and placeholder empty

BuluBulu27:
If both the textinput placeholder and value are empty, I think we should allow suggestions to be shown.

### DIFF
--- a/textinput/textinput.go
+++ b/textinput/textinput.go
@@ -824,7 +824,7 @@ func (m *Model) updateSuggestions() {
 		return
 	}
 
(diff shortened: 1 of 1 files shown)


maaslalani:
Hey @BuluBulu27, after testing this it seems that if you use dynamically set suggestions there could be a delay in the suggestions causing a flicker for a second. I would prefer if we simply require a standard placeholder so there is no delay between the placeholder showing up and the suggestions loading.

Issue #363, opened 2023-04-07
bug: table offsets too early when number of rows is less than height*2


robherley:
Using the example from the [bubbletea repo](https://github.com/charmbracelet/bubbletea/blob/5b809a3da86a63d2f07372b1941e8058178dad4d/examples/table/main.go) where the table height is `10`.

When the table is rendered with `100` rows, moving down scrolls normally and only offsetting at the last rendered row:

https://user-images.githubusercontent.com/16991201/230526308-a4ca7c4b-f8e3-4855-b7eb-7078c1bca7c3.mov

However, if we render with `15` rows (just remove the other 85 in the example), the table offsets too early on the 5th element:

https://user-images.githubusercontent.com/16991201/230526301-6dbb04f6-e856-474b-a394-614ad2a06f49.mov

For a table height of `h`, and a row length of `n`, this behavior is seen when `h < n < 2h` and will offset early on the `n - h` element

e.g. for a table height of `10` and `len(rows) == 13`, it offsets on the 3rd element

Philistino:
I was working on a table implemention with multi-select and I encountered the same problem after already implementing multi-select. My version fixes two scroll issues: the one mentioned here, and one in which a cursor move up from the last row causes an unneeded scroll up, which moves the last row out of view even when the height allows multiple rows to be shown. 

I could do a PR if it is wanted, but my current implementation has multi-select. I only noticed the scrolling issue after adding multi-select. Let me know and I can separate the multi-select and scroll fixes.

https://github.com/Philistino/bubbles/blob/master/table/table.go

maaslalani:
Fixed by #428. Thank you for the issue!

PR #429, opened 2023-11-06
fix(table): table scrolls before selected reaches top/bottom

dzeleniak:
The previous version of the code used the SetYOffset function of the table's viewport, which caused the viewport to scroll up before the selected reached the top.

Fix: #428

### DIFF
--- a/table/table.go
+++ b/table/table.go
@@ -343,9 +343,9 @@ func (m *Model) MoveDown(n int) {
 	m.UpdateViewport()
 
 	switch {
-	case m.end == len(m.rows):
+	case m.end == len(m.rows) && m.viewport.YOffset > 0:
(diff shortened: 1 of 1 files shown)


dzeleniak:
It was pointed out that a similar issue exists on scroll down. 

This only occurs in cases where `case m.end == len(m.rows)` and `case m.cursor > (m.end-m.start)/2` and len(rows) > tableHeight*2.

To fix this case I add a check to the two cases that fail to ensure they only run if the viewport's YOffset > 0.

dzeleniak:
Fix: #428 
Fix: #363

maaslalani:
Thanks so much for the PR @dzeleniak! Just tested and it works well!

Just as a heads up, we are going to swap out the entire table implementation for Lip Gloss tables.

Issue #294, opened 2022-11-16
Stream content to a viewport


oxodao:
Hi,

Let's say I'm running some kind of external software and I want to display the logs of it during its lifetime in a viewport.
How can I do this ? As far as I understand, the viewport can only be updated through the SetContent method.
Ideally, the viewport should update every new line but I don't want to do some weird things if I missed something

Minimal code:
-> /tmp/some-software.sh
```bash
while true; do echo Hey; sleep 1; done
```

```go
package main

import (
	"bufio"
	"bytes"
	"fmt"
	"io"
	"os"
	"os/exec"

	"github.com/charmbracelet/bubbles/viewport"
	tea "github.com/charmbracelet/bubbletea"
)

type CmdMsg struct {
	path string
}

type Model struct {
	logs io.Writer

	ready    bool
	viewport viewport.Model
}

func (x Model) Init() tea.Cmd {
	return func() tea.Msg {
		return CmdMsg{path: "/tmp/some-software.sh"}
	}
}

func (x Model) Update(msg tea.Msg) (tea.Model, tea.Cmd) {
	switch msg := msg.(type) {
	case tea.KeyMsg:
		if k := msg.String(); k == "ctrl+c" || k == "q" || k == "esc" {
			return x, tea.Quit
		}
	case CmdMsg:
		cmd := exec.Command("/usr/bin/bash", msg.path)

		var logs bytes.Buffer
		logsWriter := bufio.NewWriter(&logs)
		cmd.Stdout = logsWriter
		cmd.Stderr = logsWriter

		x.logs = logsWriter
		// ???

		cmd.Start()

	case tea.WindowSizeMsg:
		if !x.ready {
			x.viewport = viewport.New(msg.Width, msg.Height-10)
			x.viewport.YPosition = 5
			x.viewport.HighPerformanceRendering = false
			x.viewport.SetContent( /* ???? */ )
			x.ready = true
		} else {
			x.viewport.Width = msg.Width
			x.viewport.Height = msg.Height - 10
		}
	}

	return x, nil
}

func (x Model) View() string {
	return x.viewport.View()
}

func main() {
	p := tea.NewProgram(
		Model{},
		tea.WithAltScreen(),
	)

	if _, err := p.Run(); err != nil {
		fmt.Println("could not run program:", err)
		os.Exit(1)
	}
}
```

Thanks!

rew1nter:
I've the same question

coxley:
[...]

PR #431, opened 2023-11-09
add AppendContent and DeleteTopContent methods

ThisGuyCodes:
I have a use case where I want to stream logs to a bubbletea-based TUI, which seems like a perfect use of the `viewport.Model`! Particularly since logs can sometimes come at you quite fast, using the terminal built-in scrolling mechanisms seems prudent.

However the current `viewport.Model` only supports fully replacing the current content and re-syncing on every change.

Here I've added two methods to `viewport.Model`:
* `AppendContent`: adds to the end of the current content, and returns a tea.Cmd that "scrolls" the new content into view if necessary (i.e. if `PastBottom()` and `HighPerformanceRendering` is true).
* `DeleteTopContent`: deletes `n` lines from the top of the current content, and scrolls appropriately if this leaves you "passed the top".

Open to feedback on everything here! I'll get around to making a few test apps myself to verify behavior before taking this out of draft.

### DIFF
--- a/viewport/viewport.go
+++ b/viewport/viewport.go
@@ -108,6 +108,36 @@ func (m *Model) SetContent(s string) {
 	}
 }
 
+// AppendContent adds content to the end of the current content.
+func (m *Model) AppendContent(c []string) tea.Cmd {
+	wouldRender := m.PastBottom()
+	m.lines = append(m.lines, c...)
+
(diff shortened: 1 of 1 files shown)


rendellc:
Just wanted to chime in and say I have the same use case. Hope this gets approved

QuLogic:
This looks useful for #294, but see also #133.

PR #433, opened 2023-11-13
feat(table): add option for wraping cells

prgres:
The current implementation of the `table.renderRow` function allows only for truncating cell content, resulting in something like this: "test123 ...". 

image

My project requires full content visibility, so I need a way to handle it. My first idea was to write a custom `table` component, but I guess it may be useful to make it optional for everyone
image

I couldn't find any contribution guide, so I hope everything is ok.

### DIFF
--- a/table/table.go
+++ b/table/table.go
@@ -8,6 +8,7 @@ import (
 	tea "github.com/charmbracelet/bubbletea"
 	"github.com/charmbracelet/lipgloss"
 	"github.com/mattn/go-runewidth"
+	"github.com/muesli/reflow/wordwrap"
 )
 
 // Model defines a state for the table widget.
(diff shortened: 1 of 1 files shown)


prgres:
In my free time, I will incorporate reflow and check the outcome

prgres:
It looks ok:
image

But it messes up with movement. My assumptions lay in the viewport.YOffset but I have a hard time understanding what is going on in MoveDown and MoveUp methods

maaslalani:
Hey @prgres, thanks so much for the PR. Just as a heads up the `table` bubble will be entirely replaced implementation-wise with the Lip Gloss `table` rendering package.

Thank you for the PR again, I will close it for now while we swap out all the internals for the new renderer which will support line-wrapping and multiline rows.

image

PR #434, opened 2023-11-13
fix(table): substract headers Y size from the total table size

prgres:
I spotted a bug while creating a table and viewport side by side. Even though they have set the exact height value, both have different sizes. While debugging, it came out that we only passed it down to the nested viewport component by setting table height, but the render method looks like this.

```
// View renders the component.
	return m.headersView() + "\n" + m.viewport.View()
}
```

So, the total height of the table is the sum of the viewport's height and the headers' height. I have simply subtracted headings Y size from the viewport height, and it looks like this fixed the issue.

### DIFF
--- a/table/table.go
+++ b/table/table.go
@@ -296,8 +296,8 @@ func (m *Model) SetWidth(w int) {
 
 // SetHeight sets the height of the viewport of the table.
 func (m *Model) SetHeight(h int) {
-	_, y := m.styles.Header.GetFrameSize()
-	m.viewport.Height = h - y
+	headerHeight := lipgloss.Height(m.headersView()) // Since headers are truncated, it is always 1
+	m.viewport.Height = h - headerHeight - m.styles.Header.GetVerticalFrameSize()
 	m.UpdateViewport()
 }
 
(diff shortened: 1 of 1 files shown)


prgres:
Okay, it does not work as I expected. I will dig it up latter

prgres:
Okay, first of all, I changed `GetFrameSize` to `GetVerticalFrameSize` since X calculations are not helpful here.

Secondly, I realized that `GetVerticalFrameSize` does not return the total height of the component - just the sum of margins, padding, and borders. So, besides subtracting that one bad boi, I have to get in content height. I have used `lipgloss.Height(m.headersView())` in the `SetHeight` function. I am unsure if this approach is good since it renders it twice (in SetHeight and View functions).

What is worth mentioning is that the content height, with the current implementation, is always 0 or 1 due to truncating the header text. But I keep in mind that if https://github.com/charmbracelet/bubbles/pull/433 will merge, I would also like to make a similar one for that part. That's why I decided to calculate that value here.

prgres:
Update once again:

During testing on a local setup, it came out that `GetVerticalFrameSize` is not necessary at all. But would great if somebody could confirm that

prgres:
@meowgorithm  @muesli i do not who to ping

maaslalani:
Hey @prgres, thanks so much for the PR ❤️

maaslalani:
Thanks for the PR @prgres, I merged this one since it's a simple bug fix but just to give you a heads up, we're going to swap out the entire implementation for the Lip Gloss table renderer:

Issue #237, opened 2022-09-14
Timer and Stopwatch is not accurate


HAOMOAI:
For timer, it's 1s is actually 1.21s in real world, and 10s is about 12.13s.

For stopwatch, it says 10.39s, and actually 12.284s elapsed.

(Use the demo Go program, and keep the interval 1ms, unchanged)

apphireXB:
Same issue. @meowgorithm  please take a look.

muesli:
I do see an issue here: the timer & stopwatch don't use continuously running timers, but deduct the `interval` from the remaining `timeout` in their `Update` method. This doesn't account for the time spent in the `Update` method itself - or in other model's `Updates` in most cases.

fuhrmannb:
Hitting the same issue on my side.

Possible solutions are, for stopwatch (should be in the same spirit for timer):

- Store the start time instead of the total duration, and make the math for `Elapsed` method.
    ```golang
    func (m Model) Elapsed() time.Duration {
        time.Now().Sub(m.startTime)
    }
    ```
    It's more complicated to manage pause this way.
- Or add a `tickTime` field in `TickMsg` that stores the date when the tick has been created.
    ```golang
    type TickMsg struct {
        ID       int
        tickTime time.Time
    }
    ```
    Then increase duration in `TickMsg` by doing:
    ```golang
    m.d += time.Now().Sub(msg.tickTime)
    ```
    This solutions is easier to manage pause, but need to store 

I've tested the second solution, it's working but you need to store

Moreover, in both case, output shows `5.0012698s` and not `5s`.
We can either format `View()` output depending on the duration precision (`time.Seconds` should only show second precision, `time.Milliseconds` only ms precision); or/and have a new field to store precision.

PR #437, opened 2023-11-18
fix(stopwatch): timer was not accurate

fuhrmannb:
Stopwatch does not track correctly time since update process time can be longer that tick time update.

This patch adds a start time with `sT` attribute to track time of stopwatch start.

Also adapt view to have accuracy matching the tick interval. To have directly the raw duration value, use `Elapsed()` method.

Issue: #237


stopwatch/stopwatch.go
```diff
@@ -46,6 +46,7 @@ type ResetMsg struct {
 // Model for the stopwatch component.
 type Model struct {
 	d       time.Duration
+	sT      time.Time
```
caarlos0, line 49: i would name this `start` instead 🤔 

```suggestion
	start   time.Time
```

stopwatch/stopwatch.go
```diff
@@ -118,6 +119,12 @@ func (m Model) Update(msg tea.Msg) (Model, tea.Cmd) {
 		if msg.ID != m.id {
 			return m, nil
 		}
+		if msg.running && !m.running {
+			m.sT = time.Now()
```
caarlos0, line 123:
```suggestion
			m.start = time.Now()
```

stopwatch/stopwatch.go
```diff
@@ -118,6 +119,12 @@ func (m Model) Update(msg tea.Msg) (Model, tea.Cmd) {
...
+		}
+		if !msg.running && m.running {
+			m.d += time.Since(m.sT)
```
caarlos0, line 126:
```suggestion
			m.d += time.Since(m.start)
```

stopwatch/stopwatch.go
```diff
@@ -137,12 +143,15 @@ func (m Model) Update(msg tea.Msg) (Model, tea.Cmd) {
 
 // Elapsed returns the time elapsed.
 func (m Model) Elapsed() time.Duration {
+	if m.running {
+		return m.d + time.Since(m.sT)
```
caarlos0, line 147:
```suggestion
		return m.d + time.Since(m.start)
```

caarlos0 (changes requested):
hey, thanks for this!

just one comment regarding the variable name :)

PR #438, opened 2023-11-23
Add support for colors in table cells

sergiolms:
With this PR I want to address the following:
- Invisible characters cause incorrect calculation of cell width, making the table component to incorrectly print ellipsis when it shouldn't.
- When highlighting a row, if a cell had colors in it, the reset character would break the highlighting. I am removing the reset character and then highlighting each cell individually instead of the whole row, so that the highlight reset character serves as the reset character for the previously set color.
- While I was at it, I removed the `min` and `max` methods, since GoLang introduces them since v1.21

### DIFF
--- a/table/table.go
+++ b/table/table.go
@@ -2,6 +2,7 @@ package table
 
 import (
 	"strings"
+	"unicode"
 
 	"github.com/charmbracelet/bubbles/key"
 	"github.com/charmbracelet/bubbles/viewport"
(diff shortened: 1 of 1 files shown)


maaslalani:
Hey @sergiolms, thanks so much for the PR. Just as a heads up the `table` bubble will be entirely replaced implementation-wise with the Lip Gloss `table` rendering package.

Thank you for the PR again, I will close it for now while we swap out all the internals for the new renderer.

Issue #358, opened 2023-03-23
textinput: placeholder changes component width


seanbanko:
I'm seeing strange behavior using a Placeholder with the textinput bubble. Here is a minimum reproducible example:
```
package main

import (
	"os"

	"github.com/charmbracelet/bubbles/textinput"
	tea "github.com/charmbracelet/bubbletea"
	"github.com/charmbracelet/lipgloss"
)

type model struct {
	input textinput.Model
}

func main() {
	m := newModel()
	p := tea.NewProgram(m)
	if _, err := p.Run(); err != nil {
		os.Exit(1)
	}
}

func newModel() model {
	input := textinput.New()
	input.Prompt = " "
	input.Width = 5
	input.Placeholder = "123"
	input.Focus()
	return model{input: input}
}

func (m model) Init() tea.Cmd {
	return nil
}

func (m model) Update(msg tea.Msg) (tea.Model, tea.Cmd) {
	switch msg := msg.(type) {
	case tea.KeyMsg:
		switch msg.String() {
		case "ctrl+c":
			return m, tea.Quit
		}
	}
	var cmd tea.Cmd
	m.input, cmd = m.input.Update(msg)
	return m, cmd
}

func (m model) View() string {
	return lipgloss.NewStyle().Border(lipgloss.RoundedBorder()).Render(m.input.View())
}
```

Changing the length of the Placeholder string changes the way that the textinput is rendered before it receives the first character, triggering a resize after the first character is received. See examples with various Placeholders.

Placeholder = "123"
![demo-1](https://user-images.githubusercontent.com/74473367/227225452-cb80fafe-3d70-4b62-8807-1dec858fde75.gif)

Placeholder = "12345"
![demo-2](https://user-images.githubusercontent.com/74473367/227225484-69670867-2256-4582-8e0f-70fae4151458.gif)

Placeholder = "1234567890"
![demo-3](https://user-images.githubusercontent.com/74473367/227225494-71e4c74c-aa63-4af3-bf18-34eb49b71516.gif)

Placeholder = "12345 " (with trailing space)
![demo-4](https://user-images.githubusercontent.com/74473367/227225511-d0cadc58-2e8b-48cb-b2d2-333f6b6d98a5.gif)

no Placeholder
![demo-5](https://user-images.githubusercontent.com/74473367/227226560-79b4d948-9b4f-4881-aa23-92815fb3b7c5.gif)

[...]

PR #442, opened 2023-11-25
fix (textinput): do not resize based on placeholder

seanbanko:
This PR fixes this [issue](https://github.com/charmbracelet/bubbles/issues/358) regarding textinput resizing based on the placeholder. The original behavior is documented in the issue and has been modified to the following:

`Placeholder = ""`
empty-placeholder

`Placeholder="123"`
short-placeholder

`Placeholder = "1234567890"`
long-placeholder

### DIFF
--- a/textinput/textinput.go
+++ b/textinput/textinput.go
@@ -714,7 +714,11 @@ func (m Model) placeholderView() string {
 	v += m.Cursor.View()
 
 	// The rest of the placeholder text
-	v += style(string(p[1:]))
+	var spaces []rune
+	for i := 0; i < m.Width-lipgloss.Width(string(p[1:m.Width])); i++ {
+		spaces = append(spaces, ' ')
+	}
+	v += style(string(append(p[1:m.Width], spaces...)))
 
 	return m.PromptStyle.Render(m.Prompt) + v
 }


maaslalani (approved):
Thank you so much for this PR @seanbanko, we really appreciate it. And thanks for your patience with getting it merged!

hopefulTex:
This throws an error in the cases where the `Width` is set to `0` or is greater than the length of the placeholder.

Placeholder: `"7 runes"`

Width: `0`
`runtime error: slice bounds out of range [1:0]`
Width: `40`
`runtime error: slice bounds out of range [:40] with capacity 32`

seanbanko:
@hopefulTex thank you for catching this.

I had thought that my test with `Width = 5` and `Placeholder = "123"` would have tested the latter case of `Width` greater than the length of the placeholder, but I was surprised to learn today that upper bound for a valid slice index in Go is the capacity of the slice, not the length, which I think explains why `Width = 5` and `Placeholder = "123"` was ok but `Width = 40` and `Placeholder = "7 runes"` was not.

I'd be glad to open another PR to fix this unless you'd prefer to; I saw you had one open yourself that may have been more robust than my solution.

maaslalani:
Hey @hopefulTex thank you for catching that, I had overlooked before merging. I will revert this PR and fix!

hopefulTex:
@seanbanko I'll reopen mine since I snuck in an old paste maths fix for a different issue (cause I'm evil like that 😈)

PR #445, opened 2023-12-08
feat(viewport): add content style for controlling colors within viewport

Rydez:
There's a weird issue when attempting to change the colors within the viewport when the using [reflow/wrap](https://github.com/muesli/reflow).

Here's a code example:
```golang
var c1 = lipgloss.Color("111")
var c2 = lipgloss.Color("123")
var c3 = lipgloss.Color("239")

func main() {
	const name = "somebody"
	const msg = "a nice message about some things."

	const limit = 20

	nameStyle := lipgloss.NewStyle().Foreground(c1).Background(c3)
	msgStyle := lipgloss.NewStyle().Foreground(c2).Background(c3)
	wrapped := wrap.String(nameStyle.Render(name+": ")+msgStyle.Render(msg), limit)

	vp := viewport.New(20, 5)

	vpStyle := lipgloss.NewStyle().Foreground(c2).Background(c3)
	vp.Style = vpStyle

	vp.SetContent(wrapped)

	fmt.Println(vp.View())
}
```

which results in:
pr-screen-cap

However, with this PR and changing the code above like so:
```diff
-	vp.Style = vpStyle
+	vp.ContentStyle = vpStyl
```

the result is: 
fixed

I struggled to find a way to get this result without changing `viewport.go`.

### DIFF
--- a/viewport/viewport.go
+++ b/viewport/viewport.go
@@ -42,6 +42,10 @@ type Model struct {
 	// useful for setting borders, margins and padding.
 	Style lipgloss.Style
 
+	// Style applies a lipgloss style to the content in the viewport.
+	// Can be be helpful for controlling the background and foreground colors.
+	ContentStyle lipgloss.Style
+
 	// HighPerformanceRendering bypasses the normal Bubble Tea renderer to
 	// provide higher performance rendering. Most of the time the normal Bubble
 	// Tea rendering methods will suffice, but if you're passing content with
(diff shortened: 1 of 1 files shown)


caarlos0:
you can set the previously styled content into the viewport.

unless I'm missing something, this isn't needed I think 🤔 


the reflow/wrap thing is solved by using our new packages (`ansi.Cut`), and should be fixed on main.

Thanks for the PR though! Let me know if my assessment is incorrect :)

PR #451, opened 2023-12-23
fix(textinput): Placeholder No Longer Changes Width + Paste Calculation

hopefulTex:
Previously, the `placeholderView()` function did not respect `Width` when it was set to a non-zero value.  The textinput's width now stays the same irrelevant of the placeholder length.
Pasting now trims to match the available space if length is greater than the `charLimit`.

### DIFF
--- a/textinput/textinput.go
+++ b/textinput/textinput.go
@@ -1,6 +1,7 @@
 package textinput
 
 import (
+	"reflect"
 	"strings"
 	"time"
 	"unicode"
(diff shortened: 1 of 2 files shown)


maaslalani:
Thank you so much @hopefulTex, just tested this out with various values for placeholder and width and it works great! Really appreciate the PR ❤️

textinput/textinput.go
```diff
@@ -713,8 +713,30 @@ func (m Model) placeholderView() string {
...
+
+		// if width < len, 'subtract'(add) number to len and dont add padding
+		if availWidth < 0 {
+			minWidth += availWidth
+			availWidth = 0
+		}
+		// append placeholder[len] - cursor, append padding
+		v += style(string(p[1:minWidth]))
+		v += style(strings.Repeat(" ", availWidth))
+
```
maaslalani, line 735:
```suggestion
```

maaslalani:
> Pasting now trims to match the available space if length is greater than the charLimit.

Great catch on this as well!

PR #455, opened 2024-01-15
Update README.md with fixing outdated wrong example

yuseferi:
There was an outdated link in the readme file that led users to outdated with syntax error example. in this PR I fixed the link to refer to the correct and latest code sample.


maaslalani:
Thanks for the contribution @yuseferi, this was fixed in https://github.com/charmbracelet/bubbles/pull/467

PR #458, opened 2024-01-17
Fix list pagination functions

nekopy:
1. Set receivers for list.NextPage and list.PrevPage to pointers

# Why?
The Next and Prev list functions should advance or move back the page, but with the receivers not being pointers, it causes these methods to be no-ops

### DIFF
--- a/list/list.go
+++ b/list/list.go
@@ -532,12 +532,12 @@ func (m *Model) CursorDown() {
 }
 
 // PrevPage moves to the previous page, if available.
-func (m Model) PrevPage() {
+func (m *Model) PrevPage() {
 	m.Paginator.PrevPage()
 }
 
 // NextPage moves to the next page, if available.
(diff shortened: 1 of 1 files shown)


maaslalani:
Yes, looking at how the `CursorUp` and `CursorDown` methods have pointer receivers I believe you're correct on this one.

mikelorant:
@maaslalani @nekopy This has caused a regression for my app causing `NextPage` and `PrevPage` to jump two pages at a time.

Any ideas what is going on?

mikelorant:
Figured this out. Was clearly a bug in my app but since this was a noop, it wasn't doing anything before.

maaslalani:
> Figured this out. Was clearly a bug in my app but since this was a noop, it wasn't doing anything before.
> 
> 

Thanks for double checking @mikelorant! Let me know if there's any other problems you're running into.

Issue #464, opened 2024-01-25
table: honor a column width set to 0


fabio42:
**Describe the bug**
A column set with a `Width = 0` is rendered with the tail character `…`. While there may be a rationale for this implementation, I would like to propose a change. Currently a column width set to `0` actually has a width of `1`, which prevents the implementation of certain behaviours, such as intentionally hiding a column or dynamically removing a column when the terminal size is too small.

**Setup**
 - OS: Linux, OSX
 - Shell: zsh
 - Terminal Emulator: alacritty
 - Terminal Multiplexer: tmux
 - Locale: en_CA.UTF-8

**Source Code**
 - Create a column set with a Width = 0.
 - Observe the rendering of the column set, noting the presence of the tail character `…`.

**Expected behavior**
When the column's width is set to 0, the column should not be rendered.

**Screenshots**
<img width="283" alt="image" src="https://github.com/charmbracelet/bubbles/assets/3433725/ff87adf1-9d39-42ba-ac51-ff7928e4babe">

fabio42:
This is a proposed patch that would address this issue, please let me know what you think about it.
Thank you :)

maaslalani:
I agree with this proposal, we should assume that the user knows what they are doing if they set the width to 0, they can intentionally set the width to 1 themselves if they want to show the ellipsis!

PR #465, opened 2024-01-25
(fix)table: don't render a column with a width <= 0

fabio42:
This address #464


maaslalani:
Before:

image


After:

image

PR #473, opened 2024-02-02
fix(textinput): out of range panic if no matched suggestions

rdnt:
Fixes a possible panic in case CurrentSuggestion is called on an empty input.

### DIFF
--- a/textinput/textinput.go
+++ b/textinput/textinput.go
@@ -830,6 +830,10 @@ func (m *Model) AvailableSuggestions() []string {
 
 // CurrentSuggestion returns the currently selected suggestion.
 func (m *Model) CurrentSuggestion() string {
+	if len(m.matchedSuggestions) == 0 {
+		return ""
+	}
+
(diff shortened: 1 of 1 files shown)


textinput/textinput.go
```diff
@@ -830,6 +830,10 @@ func (m *Model) AvailableSuggestions() []string {
 
 // CurrentSuggestion returns the currently selected suggestion.
 func (m *Model) CurrentSuggestion() string {
+	if len(m.matchedSuggestions) == 0 {
```
maaslalani, line 833:
```suggestion
	if m.currentSuggestionIndex >= len(m.matchedSuggestions) {
	}
```

maaslalani:
Hey @rdnt, nice catch. Instead of checking if the `len` of the array is zero we can check if the index >= the length which will catch this empty case and other cases where it may panic.

rdnt:
Hi! :D Committed the suggestion, thanks!

PR #474, opened 2024-02-04
feat(textinput): `Yank` function that returns and resets the content

daviddenhaan:
Introduce a new `Yank` method on the textinput `Model` that first grabs the content, then resets the text input and returns the content.

### DIFF
--- a/textinput/textinput.go
+++ b/textinput/textinput.go
@@ -211,6 +211,13 @@ func (m Model) Value() string {
 	return string(m.value)
 }
 
+// Yank returns the current value of the text input, and also resets it.
+func (m *Model) Yank() string {
+	v := m.Value()
+	m.Reset()
+	return v
+}
+
 // Position returns the cursor position.
 func (m Model) Position() int {
 	return m.pos


maaslalani:
Hey @daviddenhaan, thank you for the contribution. Personally, I don't think this method is necessary since:

1. It increases the API surface area
1. it is fairly trivial to `Reset` the textinput after getting the value
1. I don't think `Yank` is the correct naming since it is ambiguous as in Vim yanking does not reset the text (a more appropriate name could be `GetValueAndReset` but then it's easier (and more clear) to: `m.Value()` and `m.Reset()`.

maaslalani:
Hey @daviddenhaan, thanks again for this PR. We really appreciate the contribution! I think for us it's better to keep the API surface area as small as possible. I agree that this method is useful but I think it's more clear for a reader to read both functions happening explicitly when written in code, instead of having to go to the definition if they are unfamiliar with the function.

That being said I think we can close this PR out for now and if it is very requested add it in.

Thank you again for the contributions and ideas!

Issue #307, opened 2023-01-08
textinput: Prompt length not taken into account for width


mikelorant:
When creating a textinput, the prompt width is added to the width of the entire textinput. This means that two text inputs with a width of 20, will have different lengths based on the length of the prompt. This is visible when adding a border.

```
func initialModel() model {
	tiA := makeTextInput("1234")
	tiB := makeTextInput("1234567890")

	return model{
		textInputA: tiA,
		textInputB: tiB,
	}
}

func makeTextInput(str string) textinput.Model {
	ti := textinput.New()
	ti.Prompt = str
	ti.Focus()
	ti.CharLimit = 156
	ti.Width = 20

	return ti
}
```

![](https://d.pr/i/0R02PO.png)

I have an example of this bug here:
https://github.com/mikelorant/bubbles-textinput-prompt

Looking at the textinput code, it is clear that the prompt length is never subtracted from the width. I believe the width is actually shorter than it should be but I am unclear what the intended width is meant to represent.

mikelorant:
@meowgorithm Can we get some clarification of what the width is meant to mean. You were the last person to touch this field so I am hoping you know the reasons behind why 😄 

```
// Width is the maximum number of characters that can be displayed at once.
// It essentially treats the text field like a horizontally scrolling
// viewport. If 0 or less this setting is ignored.
Width int
```
https://github.com/charmbracelet/bubbles/blob/master/textinput/textinput.go#L113-L116

I have two possible interpretations:
1. The width is the entire `textinput` component.
2. The width is just the visible area for user input space. Promps and extra padding ADD to this width.

Whatever the answer is, I believe we should update the documentation clarifying this field. If there is a bug, I will put up the appropriate pull request to fix this.

meowgorithm:
Okay, jumping back in on this one: `Width` here is the width of the `textinput` input area only — so not including margins, borders, padding and so on.

mikelorant:
@meowgorithm Thanks for the clarification. Will close this issue as it seems everything is working as expected.

mikelorant:
@meowgorithm Thanks for the clarification. Will close this issue as it seems everything is working as expected.

mikelorant:
@meowgorithm Thanks for the clarification. Will close this issue as it seems everything is working as expected.

PR #484, opened 2024-02-28
fix: handle textinput max-width edge cases

### LINKED ISSUE #358: textinput: placeholder changes component width

JaredReisinger:
I noticed that the cursor seemed to be adding an extra space which the width calculations weren't accounting for. This also led me to discover an edge case with the view (`offset`/`offsetRight`) calculation when the cursor (`pos`) sits at the very end. Additionally, `View()` was always rendering the cursor, even when `focus` was false, which the comment suggests should not be the case.

These should all be resolved, such that `Width` is now fully respected in both `View()` and `placeholderView()`: they should _never_ render a string that exceeds the requested width. (Well... with one caveat: I did not touch the suggestion logic, as I've never used it. I did leave a comment, however, that perhaps suggestions shouldn't even render when the component does not have focus.)

Fixes #358, #307

### DIFF
--- a/textinput/textinput.go
+++ b/textinput/textinput.go
@@ -358,20 +358,28 @@ func (m *Model) handleOverflow() {
 
 		m.offsetRight = m.offset + i
 	} else if m.pos >= m.offsetRight {
-		m.offsetRight = m.pos
+		// we need to show the character *at* m.pos, if there is one
+		m.offsetRight = min(m.pos+1, len(m.value))
 
 		w := 0
+
+		// if m.pos is *past* m.offsetRight, that's one character of width that
+		// we can't display from m.value
+		if m.pos > m.offsetRight {
+			w++
+		}
+
 		runes := m.value[:m.offsetRight]
-		i := len(runes) - 1
+		i := len(runes) // 'i' represents the last rune we *can* fit
 
 		for i > 0 && w < m.Width {
-			w += rw.RuneWidth(runes[i])
+			w += rw.RuneWidth(runes[i-1])
 			if w <= m.Width {
 				i--
 			}
 		}
 
-		m.offset = m.offsetRight - (len(runes) - 1 - i)
+		m.offset = m.offsetRight - (len(runes) - i)
 	}
 }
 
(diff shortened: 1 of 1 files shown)


JaredReisinger:
My particular use-case involves dynamically setting the `Width` _as_ edits are happening (it's a semi-WYSIWYG-style TUI), using `len(text.Value()) + 1` when it has focus, and `len(text.Value())` when it does not.  Without this PR, the textinput was always rendering two extra spaces at the end.  With this fix, the textinput only renders the requested `Width`.

I can provide a minimal repro example if needed.

maaslalani:
@JaredReisinger A minimal reproducible example would be great, showing the error / issue and then how this PR fixes that!

PR #490, opened 2024-02-29
test(textarea): Improve textarea tests

mikelorant:
Improve textarea tests to check the cursor location. This is the displayed location and not the line row and column (which does not take into account soft wrapping).

As part of this improvement, a number of existing tests were merged into the view test tables as it was unnecessary duplication of code having them separated.

### DIFF
--- a/textarea/textarea_test.go
+++ b/textarea/textarea_test.go
@@ -10,101 +10,6 @@ import (
 	tea "github.com/charmbracelet/bubbletea"
 )
 
-func TestNew(t *testing.T) {
-	textarea := newTextArea()
-	view := textarea.View()
-
-	if !strings.Contains(view, ">") {
-		t.Log(view)
-		t.Error("Text area did not render the prompt")
-	}
-
-	if !strings.Contains(view, "World!") {
-		t.Log(view)
-		t.Error("Text area did not render the placeholder")
-	}
-}
-
-func TestInput(t *testing.T) {
-	textarea := newTextArea()
-
-	input := "foo"
-
-	for _, k := range []rune(input) {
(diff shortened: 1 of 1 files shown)


mikelorant:
@maaslalani Converting the older tests into the test tables.

textarea/textarea_test.go
```diff
@@ -424,6 +329,8 @@ func TestView(t *testing.T) {
 		name      string
 		modelFunc func(Model) Model
 		expected  string
```
maaslalani, line 331: If it's not too much work I would change expected to be:

```go
struct {
    value string
    cursorRow int
    cursorCol int
}
```

mikelorant: Will do, makes much more  sense.

Any other changes you'd like?

maaslalani: That's all, thank you for this PR!

mikelorant: Decided to go with a common convention of putting everything in the struct `want`. I also decided to go `view` instead of `value` as that is actually what is stored in that string.

mikelorant:
My intention after this gets merged is to also add checks for the value so we can confirm what is provided back to the user. What we render is not the same as what the value might be due to wrapping and scrolling.

mikelorant:
@meowgorithm @maaslalani This one is also ready to be review. Just improves the tests, no changes to any other functionality.

maaslalani:
Thanks for this one @mikelorant, really appreciate the well written tests!

mikelorant:
No problem. I'll expand the tests to check a few more things and we should be able to convert nearly every other test into a single test table.

PR #491, opened 2024-02-29
fix(textarea): Fix end of buffer character

mikelorant:
The end of buffer character was not being rendered when line numbers were not enabled.

When the new test tables unit tests were created this issue was identified however to limit the scope of those changes to just creating unit tests, this regression was maintained.

This change fixes this issue and updates the tests to the intended result.

### DIFF
--- a/textarea/textarea.go
+++ b/textarea/textarea.go
@@ -1138,10 +1138,7 @@ func (m Model) View() string {
 		s.WriteString(prompt)
 		displayLine++
 
-		if m.ShowLineNumbers {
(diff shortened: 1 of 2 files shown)


maaslalani:
Awesome, thanks @mikelorant! 

Going to merge this since it is the way it works in vim but just as a heads up may revert this one in the future depending on what users expect.

mikelorant:
Totally understand if you want to revert functionality.

My personal view is that end of buffer character should be set to a space character which would make it invisible. The developer should have to set it to make it available.

maaslalani:
Sounds good, I agree on the setting to space character as well (for now)! Thanks for all the awesome work @mikelorant!

Issue #483, opened 2024-02-24
textinput: add support for wrapping the prompt


twpayne:
**Is your feature request related to a problem? Please describe.**

My application uses `textinput`s with long prompts. The prompts are long because they contain multiple filenames from the user that can get quite long, for example:

```console
$ chezmoi rm ~/.config/git/config
Remove /Users/twp/.config/git/config and /Users/twp/.local/share/chezmoi/home/private_dot_config/git/config.tmpl? yes/no/all/quit
```

`textinput` only displays the prompt up to the width of the terminal, which leads to the prompt being truncated and confusing, for example, with a terminal of width 80:

```console
$ chezmoi rm ~/.config/git/config
Remove /Users/twp/.config/git/config and /Users/twp/.local/share/chezmoi/home/pr
```

**Describe the solution you'd like**

I'd like a boolean option to wrap the prompt over multiple lines, wrapping words (maybe using github.com/muesli/reflow) so you would get the following with a terminal width of 80:

```console
$ chezmoi rm ~/.config/git/config
Remove /Users/twp/.config/git/config and /Users/twp/.local/share/chezmoi/home/
private_dot_config/git/config.tmpl? yes/no/all/quit
```

**Describe alternatives you've considered**

Forking `textinput` to add this functionality.

**Additional context**

#417 talks about multiline **text**, this feature request is about a multiline **prompt**.

I would happily submit a PR for this if you are interested.

meowgorithm:
This all makes sense to me; thanks for the well written writeup, Tom. If this is something you'd like to work on, by all means go for it.

maaslalani:
For what it's worth, `textarea` can support multiline _prompts_ with this PR:

* https://github.com/charmbracelet/bubbles/pull/302

meowgorithm:
Hey, Tom. Just following up on this one: it looks like #302 was merged. If this still an issue for you, or if the problem’s solved, would love to know.

twpayne:
I think #302 is for textarea responses. This issue is for textinput prompts, i.e. what should happen if [`len(github.com/charmbracelet/bubbles/textinput.Model.Prompt])`](https://pkg.go.dev/github.com/charmbracelet/bubbles@v0.20.0/textinput#Model) is greater than the width of the terminal?

meowgorithm:
Oh indeed, #302 is for placeholder text.

[...]

PR #492, opened 2024-03-01
fix(textinput): make textinput.Model satisfy tea.Model interface

twpayne:
An early step towards #483.

### DIFF
--- a/textinput/textinput.go
+++ b/textinput/textinput.go
@@ -176,6 +176,11 @@ func New() Model {
 // Deprecated: Use [New] instead.
 var NewModel = New
 
+// Init exists to satisfy the tea.Model interface.
+func (m Model) Init() tea.Cmd {
+	return nil
+}
+
 // SetValue sets the value of the text input.
 func (m *Model) SetValue(s string) {
 	// Clean up any special characters in the input provided by the
@@ -553,7 +558,7 @@ func (m Model) echoTransform(v string) string {
 }
 
 // Update is the Bubble Tea update loop.
-func (m Model) Update(msg tea.Msg) (Model, tea.Cmd) {
+func (m Model) Update(msg tea.Msg) (tea.Model, tea.Cmd) {
 	if !m.focus {
 		return m, nil
 	}
--- a/textinput/textinput_test.go
+++ b/textinput/textinput_test.go
@@ -1,0 +1,8 @@
+package textinput_test
+
+import (
+	"github.com/charmbracelet/bubbles/textinput"
+	tea "github.com/charmbracelet/bubbletea"
+)
+
+var _ tea.Model = &textinput.Model{}


bashbunni:
Hey @twpayne! Heads up, we're figuring out a solution to this right now. When we try to make the bubbles implement the `tea.Model` interface, we end up having to do a lot of type assertions. For example, here's what it would look like to update a `viewport` bubble that implements `tea.Model`:
```go
// Pretend viewport.Model implements
// tea.Model here
type model {
    vp viewport.Model
}

// Some time later...
newModel, cmd := m.vp.Update(msg)
if newVPModel, ok := newModel(viewport.Model); ok {
    m.vp = newCVPModel
}
```
by comparison, this is the solution with the way to update the `viewport` model right now:
```go
m.vp, cmd := m.vp.Update(msg) // DONE
```

Just wanted to be transparent on this one as any activity on this PR is likely going to be on hold until these refactors are ironed out :)

twpayne:
Ah, thanks for the info. I wanted to add this so I could easily add tests in a follow-up PR. I've been using [the following generic function](https://github.com/twpayne/chezmoi/blob/88b586dd4f6f9c61413631369f6a7195a6935c2f/internal/chezmoibubbles/chezmoibubbles_test.go#L37-L50) in my tests for [my small collection of bubbles](https://github.com/twpayne/chezmoi/tree/88b586dd4f6f9c61413631369f6a7195a6935c2f/internal/chezmoibubbles):

```go
func testRunModelWithInput[M tea.Model](t *testing.T, model M, input string) M {
	t.Helper()
	for _, msg := range makeKeyMsgs(input) {
		m, _ := model.Update(msg)
		var ok bool
		model, ok = m.(M)
		assert.True(t, ok)
	}
	return model
}
```

This allows me to write generic tests without type assertions, [like this](https://github.com/twpayne/chezmoi/blob/88b586dd4f6f9c61413631369f6a7195a6935c2f/internal/chezmoibubbles/boolinputmodel_test.go#L64-L68):

```go
		t.Run(tc.name, func(t *testing.T) {
			actualModel := testRunModelWithInput(t, NewBoolInputModel("prompt", tc.defaultValue), tc.input)
			assert.Equal(t, tc.expectedCanceled, actualModel.Canceled())
			assert.Equal(t, tc.expectedValue, actualModel.Value())
		})
```

I'm sure you've already considering this and similar approaches.

meowgorithm:
Hey, @twpayne — just wanted to follow up on this one. This is something we're planning on addressing in Bubbles v2 because it changes the API pretty fundamentally.

Your examples here are a big part of why we're open to making this change, so thank you!

twpayne:
Great, it sounds like this PR will be superseded, so I'll close it for now.

meowgorithm:
Sounds good, Tom. I'll mention you in the proposal when we have it.

PR #494, opened 2024-03-09
Fix division by zero in viewport scrollpercentage calculation

zMoooooritz:
In the current implementation one gets a division by zero in case `len(m.lines) == m.Height + 1`, which in turn makes the function return `NaN`.
Removing the `-1` does avoid this error and still yields sensible results.

### DIFF
--- a/viewport/viewport.go
+++ b/viewport/viewport.go
@@ -92,7 +92,7 @@ func (m Model) ScrollPercent() float64 {
 	}
 	y := float64(m.YOffset)
 	h := float64(m.Height)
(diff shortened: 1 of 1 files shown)


meowgorithm:
Nice, thanks for this. Just so we can verify this fix, can you share some code that reproduces the issue this solves for?

meowgorithm (approved):

meowgorithm:
Glorious; thanks for the detailed explanation and example. This patch looks good and I was able to verify the fix. Thanks for the contribution, @zMoooooritz!

PR #500, opened 2024-03-22
Fix textinput.CurrentSuggestion() panic

KevM:
When suggestions are not yet set CurrentSuggestion() will panic. This change fixes that with a guard and returns an empty string when there is no current suggestion.

I added a test to ensure the fix works.


maaslalani:
@KevM I believe this is fixed now as there is a merge conflict with both performing a `len` check. Can you confirm if this is needed / still valid?

maaslalani:
Thanks! Tests are always great, I will merge this.

PR #508, opened 2024-04-17
New custom quick fix list bubble

Genekkion:
Hi, I've created a simple quick fix window that is similar to the ones used in Neovim. It is primarily aimed towards full screen applications and can be used for simple navigation on smaller lists. It follows a similar structure to that of the regular List component that already exists, but allows for a popup based model which wraps around existing views.

### DIFF
--- a/README.md
+++ b/README.md
@@ -212,7 +212,7 @@ func (m Model) Update(msg tea.Msg) (tea.Model, tea.Cmd) {
   customization options.
 * [evertras/bubble-table](https://github.com/Evertras/bubble-table): Interactive,
   customizable, paginated tables.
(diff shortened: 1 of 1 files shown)


maaslalani:
Hey @Genekkion, this seems like a fantastic tool! Awesome work. I'd love to add it to the Charm in the wild list. The only thing I would request is that the name is super close to one of our other tools called [`gum`](https://github.com/charmbracelet/gum) (which is actually named after Bubble Gum) so to avoid confusion I was wondering if you'd be open to renaming the tool so we can get it on the "In the Wild" list!

Genekkion:
Hi! I have renamed the project to "theHermit" as part of my series of projects!
The repo can now be found [here](https://github.com/Genekkion/theHermit).

PR #510, opened 2024-04-21
fix: default textarea EndOfBufferCharacter should be ' '

blvrd:
This changes the default `EndOfBufferCharacter` to a space character. I tend to agree with the conclusion in #491 that this is probably a better default.

Happy to discuss if this isn't the direction you want to go @maaslalani.


mikelorant:
I agree with you (obviously) that this should be the default. Be aware that there was a desire to emulate the look of Vim (something I learned later from my discussions with @maaslalani).

mikelorant (approved):

maaslalani:
Thank you so much for this one! This is definitely the right default, I believe!

PR #512, opened 2024-04-25
Fixed cursor not showing underlying char when scrolled to right

mfulz:
When scrolling to the right with limited Width the Cursor will be set to " " (space) even if there are still characters underneath.

This commit fixes this issue.

### DIFF
--- a/textinput/textinput.go
+++ b/textinput/textinput.go
@@ -17,8 +17,10 @@ import (
 )
 
 // Internal messages for clipboard operations.
-type pasteMsg string
-type pasteErrMsg struct{ error }
+type (
+	pasteMsg    string
+	pasteErrMsg struct{ error }
+)
 
 // EchoMode sets the input behavior of the text input field.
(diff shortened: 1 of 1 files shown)


mfulz:
Added logic to calculate width including cursor (if at the end) #484

bashbunni:
Hey @mfulz could you please provide some example output of what the behaviour was before compared to after this change? Also, an example or steps to test this change would also allow us to review this more quickly. Thank you :)

mfulz:
> Hey @mfulz could you please provide some example output of what the behaviour was before compared to after this change? Also, an example or steps to test this change would also allow us to review this more quickly. Thank you :)

Sure here is the issue with the actual textinput where you can't see the last char "g" under the cursor, if textinput is smaller than the text and scrolled in from the right:
orig

And here is the fixed version from this commit:
fixed

You can check that out with the textinput example from bubbletea and just reduce the ti.Width to 5

PR #526, opened 2024-05-23
feat(ci): use goreleaser for releases

aymanbagabas:
This will publish and announce releases on tag push.

### DIFF
--- a/.goreleaser.yml
+++ b/.goreleaser.yml
@@ -1,0 +1,8 @@
+includes:
+  - from_url:
+      url: charmbracelet/meta/main/goreleaser-lib.yaml
+  - from_url:
+      url: charmbracelet/meta/main/goreleaser-announce.yaml
+
+# yaml-language-server: $schema=https://goreleaser.com/static/schema-pro.json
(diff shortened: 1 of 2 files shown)


caarlos0 (approved):

meowgorithm (approved):

meowgorithm (changes requested):
Actually, would prefer if we didn't announce on push.

aymanbagabas:
> Actually, would prefer if we didn't announce on push.

It only announces patch releases, other releases don't get announced automatically.

.goreleaser.yml
```diff
@@ -0,0 +1,8 @@
+includes:
+  - from_url:
+      url: charmbracelet/meta/main/goreleaser-lib.yaml
+  - from_url:
+      url: charmbracelet/meta/main/goreleaser-announce.yaml
```
aymanbagabas, line 5: I removed this from other projects

caarlos0: done on ff5fb57

Issue #479, opened 2024-02-09
Viewport may not scroll down to the end of large text inputs


SineMah:
Viewport may not scroll down to the end of large text inputs. It will work if you reduce viewport height to e.g. 5. 
```m.contentView = viewport.New(msg.Width, 5)```

Use up and down arrows in the example.
The text is cropped after a you reach the end with arrow down. `m.conetntView.ScrollPercent` results in 100.

Setup
* OS macOS Sonoma 14.3
* zsh
* item2

Example Code
```go
package main

import (
	"encoding/json"
	"fmt"
	"github.com/charmbracelet/bubbles/viewport"
	tea "github.com/charmbracelet/bubbletea"
	"os"
)

type mdModel struct {
	ready         bool
	contentView   viewport.Model
	formatContent string
}

[...]

PR #530, opened 2024-05-29
fix(viewport): viewport may not scroll down to the end when content exceed the viewport.Width

pikomonde:
fix https://github.com/charmbracelet/bubbletea/issues/1017
fix https://github.com/charmbracelet/bubbles/issues/479


caarlos0:
this should be fixed on main.

thanks for the PR!

Issue #531, opened 2024-05-30
textinput: when `Width` is larger than the capacity of the placeholder, it causes a panic


MikaelFangel:
**Describe the bug**
When you increase the width above 31 and the placeholder capacity is set to 32, the program panics. This bug is causing the bug reported here https://github.com/charmbracelet/gum/issues/591 as far as I can tell. 

**To Reproduce**
1. Run the code with a `ti.Width` of 31
2. Run the code and see it works just fine
3. Change the value to 32
4. Run the code a see the out-of-bounds panic

**Source Code**
```go
package main

import (
	"fmt"
	"log"

	"github.com/charmbracelet/bubbles/textinput"
	tea "github.com/charmbracelet/bubbletea"
)

func main() {
	p := tea.NewProgram(initialModel())
	if _, err := p.Run(); err != nil {
		log.Fatal(err)
	}
}

type model struct {
	textInput textinput.Model
	err       error
}

func initialModel() model {
	ti := textinput.New()
	ti.Placeholder = "作業ディレクトリを指定してください"
	ti.Width = 32

	return model{
		textInput: ti,
		err:       nil,
	}
}

func (m model) Init() tea.Cmd {
	return textinput.Blink
}

func (m model) Update(msg tea.Msg) (tea.Model, tea.Cmd) {
	return m, nil
}

func (m model) View() string {
	return fmt.Sprintf(
		m.textInput.View(),
	) + "\n"
}
```

**Expected behavior**
The placeholder capacity should not cause out-of-bounds panics based on the `ti.Width`.

**Screenshots**
`p` getting a default capacity of 32
![20240530_11h39m01s_grim](https://github.com/charmbracelet/bubbles/assets/34864484/3db9fac7-8384-48ba-ac68-561bad4a18ca)

The `minWidth` exceeds the capacity
![20240530_11h40m06s_grim](https://github.com/charmbracelet/bubbles/assets/34864484/0f995c60-fb96-4ea9-8fc3-1f553bbb566f)

The out-of-bounds slide happening
![20240530_11h41m16s_grim](https://github.com/charmbracelet/bubbles/assets/34864484/cac552d7-04a3-4f5d-ba43-21fd1160c1d1)

PR #532, opened 2024-05-30
fix(textinput): slicing outside cap

MikaelFangel:
Fixes: #531
Fixes: https://github.com/charmbracelet/gum/issues/591

### Changes
 - Manually assigned `p` with the size of the `Width + 1` because when you use `var()` to assign, the cap defaults to 32 and thus it is possible to get out of bounds errors.

### DIFF
--- a/textinput/textinput.go
+++ b/textinput/textinput.go
@@ -17,8 +17,10 @@ import (
 )
 
 // Internal messages for clipboard operations.
-type pasteMsg string
-type pasteErrMsg struct{ error }
+type (
+	pasteMsg    string
+	pasteErrMsg struct{ error }
+)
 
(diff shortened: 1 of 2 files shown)


caarlos0 (approved):
thanks! can you add a test for it?

MikaelFangel:
I appreciate your patience with this.
I've added a test that makes slicing outside cap occur when the fix is not applied. (:

MikaelFangel:
Sure, I've added a test now that fails without the fix applied.

MikaelFangel:
@caarlos0 - Do I need to do anything else before it can be merged? 😄

bashbunni:
@MikaelFangel thank you so much for identifying this and fixing it!

PR #540, opened 2024-06-17
fix: progress spring defaults should't override WithStringOptions

nervo:
In progress component constructor, default spring option must be handled *after*  applying options :)

Indeed, the `springCustomized` variable is set by the option `WithSpringOptions`.


meowgorithm:
Good catch. Thanks @nervo!

PR #541, opened 2024-06-21
fix: make DefaultDelegate satisfy ItemDelegate interface

bashbunni:


maaslalani (approved):

list/defaultitem.go
```diff
@@ -127,7 +127,7 @@ func (d DefaultDelegate) Spacing() int {
 }
 
 // Update checks whether the delegate's UpdateFunc is set and calls it.
-func (d DefaultDelegate) Update(msg tea.Msg, m *Model) tea.Cmd {
+func (d DefaultDelegate) Update(ctx tea.Context, msg tea.Msg, m *Model) tea.Cmd {
```
maaslalani, line 130: Does this not need to return the model?

bashbunni: It seems the `ItemDelegate` interface has a signature `Update(ctx tea.Context, msg tea.Msg, m *Model) tea.Cmd`... So not sure if the `ItemDelegate` should be the same Update sig as a bubble tea model

https://github.com/charmbracelet/bubbles/blob/64a67d167062e075d80a132afc0851fd1b2c6b89/list/list.go#L51

bashbunni:
Hey @aymanbagabas do you mind taking a look at this when you have a moment?

aymanbagabas (approved):

PR #543, opened 2024-06-24
feat: make progress use lipgloss styles

nervo:
Progress component is not cleanly usable in a lipgloss style driven project, because of its dependencies on termenv colors and color profile. Indeed, Both of them are not directly exposed by lipgloss styles..

Btw, i've managed to make it use lipgloss styles in a backward compatible way \o/

- new `FullStyle` and `EmtpyStyle` variables are introduced, having the default foreground color of `FullColor` and `EmptyColor`
- `FullColor` and `EmptyColor` are deprecated, make their default values to empty string. This way, if they are defined, we can set the foreground during rendering.
- `WithSolidFill` now just apply foreground color to `FullStyle`
- introduce `WithFillStyles` option which set both `FullStyle` and `EmtpyStyle`. It was inspired by `WithFillCharacters`, but can be easily split in both `WithFullStyle` and `WithEmtpyStyle` if you prefer
- `WithColorProfile` is the most touchy part :) It recreates and make them inherit from themselves `FullStyle` and `EmtpyStyle`, but starting from the default renderer, and applies color profile to it

The rest is all about using `FullStyle` and `EmtpyStyle` to render runes, instead of termenv.String.

Happy comments :)

### DIFF
--- a/progress/progress.go
+++ b/progress/progress.go
@@ -10,7 +10,6 @@ import (
 	tea "github.com/charmbracelet/bubbletea"
 	"github.com/charmbracelet/harmonica"
 	"github.com/charmbracelet/lipgloss"
-	"github.com/charmbracelet/x/ansi"
 	"github.com/lucasb-eyer/go-colorful"
 	"github.com/muesli/termenv"
 )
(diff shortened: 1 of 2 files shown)


meowgorithm:
Hey again, @nervo! Nicely done here. We'll review this in a little more detail but generally speaking this looks awesome and is much needed, so thank you!

polothy:
Thanks for the tip!

PR #554, opened 2024-07-14
fix(help): wrong full help sep rendering

luevano:
No matter the available horizontal space, sometimes the separator for the full help is not rendered. For example:

image
image

The issue is that the check if a separator needs to be rendered is done on `group` (the column) and not `groups` (the total columns). With this fixed:

image
image

However, when there is enough space for the separator but not the column itself it renders the separator at the end:
image

This PR fixes these 2 issues and matches the behavior of the short help (adding a tail if needed/possible, rendering the separator inline, etc.). Behavior now:
image
image
image

### DIFF
--- a/help/help.go
+++ b/help/help.go
@@ -124,28 +124,23 @@ func (m Model) ShortHelpView(bindings []key.Binding) string {
 			continue
 		}
 
+		// Sep
 		var sep string
 		if totalWidth > 0 && i < len(bindings) {
 			sep = separator
 		}
 
+		// Item
 		str := sep +
 			m.Styles.ShortKey.Inline(true).Render(kb.Help().Key) + " " +
 			m.Styles.ShortDesc.Inline(true).Render(kb.Help().Desc)
-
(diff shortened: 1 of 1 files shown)


caarlos0 (approved):

bashbunni:
We should add a test for this

meowgorithm:
So I think this makes sense. @luevano, do you mind providing some minimal code we can use to replicate this (in addition to a test)?

meowgorithm (changes requested):
Alright, this is a solid fix. It just needs a test and we're good to go.

meowgorithm:
Alright, I have a test for this which I'll merge next. Thanks again for the excellent PR, and for your great patience with the one, @luevano.

luevano:
Hey @meowgorithm, sorry forgot about the request for the minimal code and test, been busy with work stuff lately; thanks for keeping things moving here for me!

meowgorithm:
No worries, thanks again for the great PR!

PR #556, opened 2024-07-17
feat(textinput): expose matched suggestions and index

luevano:
In its current state, if I want to get the matched suggestions (and their index) I need to recalculate that myself when the bubble already does this for free (as in already available). This allows to use the matching suggestions data to display any sort of custom views.

For example, in its current state, if I want to list the suggestions I only have access to the `Available` ones, and this is all I can display (without recalculating myself):
image
image

With these exposed methods, now I can display more "complex" views, such as:
image

Pardon the styling, just a placeholder.

### DIFF
--- a/textinput/textinput.go
+++ b/textinput/textinput.go
@@ -813,14 +813,27 @@ func (m Model) completionView(offset int) string {
 	return ""
 }
 
-// AvailableSuggestions returns the list of available suggestions.
-func (m *Model) AvailableSuggestions() []string {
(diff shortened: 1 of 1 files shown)


meowgorithm:
Good call; I believe this makes sense.

Fwiw, I think there are a lot of improvements that can happen to suggestions in general and I've been wondering if the API needs to be reexamined altogether.

caarlos0 (approved):

meowgorithm:
This is a good one. Thanks for the contribution, @luevano.

PR #560, opened 2024-07-18
fix(textinput): wrong placeholder/completion rendering

luevano:
Fixes:
- Completion using Placeholder style (only the cursor was using the Completion style)
- Extra padding after completion.
- Wrong completion char under cursor when overflowing (using index of viewable viewport instead of the completion)

TODO/discuss:
 - What should the Width apply to? both the input text + cursor? only the text? the whole thing with the prompt? Because as of right now, the actual width of the view is Width + cursor + prompt

Some examples, using contrasting colors for easier identification. Showing a half box below just to see the width, and appended a "Ñ" to see the end of the input view.

**Before fix:**
Width set to 10, showing the placeholder color:
image
Extra padding after completion (what should be appended without completion), and the color under the cursor is what should be displayed for the whole completion:
image
Wrong completion character showing under cursor when overflowing (should show a 'k'):
image

After fix:
Correct completion style applied:
image
Correct completion char used under cursor when overflowing:
image

### DIFF
--- a/textinput/textinput.go
+++ b/textinput/textinput.go
@@ -647,31 +647,48 @@ func (m Model) Update(msg tea.Msg) (Model, tea.Cmd) {
 
 // View renders the textinput in its current state.
 func (m Model) View() string {
-	// Placeholder text
-	if len(m.value) == 0 && m.Placeholder != "" {
-		return m.placeholderView()
+	var (
+		p          = m.PromptStyle.Render(m.Prompt)
+		styleText  = m.TextStyle.Inline(true).Render
+		availWidth = m.Width
+	)
+
+	// No input
+	if len(m.value) == 0 {
+		return p + m.placeholderView()
 	}
 
-	styleText := m.TextStyle.Inline(true).Render
-
-	value := m.value[m.offset:m.offsetRight]
+	// Current cursor position in viewport
 	pos := max(0, m.pos-m.offset)
-	v := styleText(m.echoTransform(string(value[:pos])))
 
+	// Render viewable input
+	value := m.value[m.offset:m.offsetRight]
+	v := styleText(m.echoTransform(string(value[:pos])))
+	availWidth -= uniseg.StringWidth(string(value[:pos]))
+
+	// Still have input to render (if the cursor is going back)
 	if pos < len(value) {
-		char := m.echoTransform(string(value[pos]))
-		m.Cursor.SetChar(char)
-		v += m.Cursor.View()                                   // cursor and text under it
-		v += styleText(m.echoTransform(string(value[pos+1:]))) // text after cursor
-		v += m.completionView(0)                               // suggested completion
+		m.Cursor.SetChar(m.echoTransform(string(value[pos])))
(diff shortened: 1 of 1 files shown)


luevano:
> Thanks, @luevano. Could you provide some code we can use to see the before and after?

Hi @meowgorithm, no problem, made it as short as possible (`ctrl+o` to toggle size and `ctrl+c` to quit):

```go
package main

import (
	"github.com/charmbracelet/bubbles/key"
)

type Model struct {
	input textinput.Model
	long  bool

	size,
	sizeShort int
	// pre-rendered
	end string
}

func newModel() *Model {
	size, sizeShort := 64, 10
	input := textinput.New()
	input.Placeholder = "Long placeholder..."
	input.ShowSuggestions = true
	input.Width = size // init with long width
	input.PlaceholderStyle = lipgloss.NewStyle().Foreground(lipgloss.Color("#EB5E28"))
	input.CompletionStyle = lipgloss.NewStyle().Foreground(lipgloss.Color("#008080"))

	input.SetSuggestions([]string{
		"hello",
		"world",
		"some long suggestion",
		"another very long suggestion...",
	})

	return &Model{
		input:     input,
		long:      true,
		size:      size,
		sizeShort: sizeShort,
		end:       lipgloss.NewStyle().Foreground(lipgloss.Color("#FF0000")).Render("Ñ"),
	}
}

func (m *Model) Init() tea.Cmd {
	m.input.CursorEnd()
	return tea.Sequence(
		m.input.Focus(),
		textinput.Blink,
	)
}

func (m *Model) Update(msg tea.Msg) (tea.Model, tea.Cmd) {
	switch msg := msg.(type) {
	case tea.KeyMsg:
		switch {
		case key.Matches(msg, key.NewBinding(key.WithKeys("ctrl+c"))):
			return m, tea.Quit
		case key.Matches(msg, key.NewBinding(key.WithKeys("ctrl+o"))):
			m.long = !m.long
			if m.long {
				m.input.Width = m.size
			} else {
				m.input.Width = m.sizeShort
			}
		}
	}

	input, cmd := m.input.Update(msg)
	m.input = input
	return m, cmd
}

func (m *Model) View() string {
	return m.input.View() + m.end
}

func main() {
	if _, err := tea.NewProgram(newModel()).Run(); err != nil {
		panic(err)
	}
}
```

caarlos0 (approved):

bashbunni:
Hey @luevano sorry for the delay on this. The changes look great. I think width should include everything (even cursor + prompt) as that will be the most predictable behaviour

bashbunni:
Update, team agreed that the width should contain all components including the text field, cursor, and prompt

luevano:
@bashbunni thanks for the feedback, I agree with that.
I did, however, try to fix the whole thing to fit within the given width, fut failed miserably lol I did try to do that back when I was working on this fix, will give it another go and see what I can do.

bashbunni:
@luevano no problem! I can see about fixing the width :) thanks again for your work on this

Issue #550, opened 2024-07-11
List `Model` `Index` returns index offset based on filtered list, but `SetItem` expects global offset in items list.


blkt:
**Describe the bug**
The index value retrieved by using `Index` returns an offset that implicitly takes into account whether filtering is active, so an hypothetical value o `0` is actually the first item in the filtered list of items.

On the other hand, `SetItem` expects an index to be used with the _global_ list of items, both filtered and unfiltered.

This makes it hard (if not impossible) to update an item within a filtered list.

I definitely might be using bubbletea list wrong, please let me know if that's the case.

**Setup**
Please complete the following information along with version numbers, if applicable.
 - OS macOS
 - Shell zsh
 - Terminal Emulator iterm
 - Terminal Multiplexer none
 - Locale en_US.UTF-8

**To Reproduce**
Using code below:
1. using `/` filter by `b`
2. push `space` on the first item
4. remove filtering
5. item at position `0`, i.e. `a`, was replaced instead of `b`

**Source Code**
Here is a minimal model usable to reproduce the issue, also found [here](https://github.com/stacklok/minder/blob/main/internal/util/cli/multi_select.go#L137-L143).
```
package main

import (
	"cmp"
	"fmt"
	"io"
	"slices"
	"strings"

	"github.com/charmbracelet/bubbles/key"
	"github.com/charmbracelet/bubbles/list"
	tea "github.com/charmbracelet/bubbletea"
	"github.com/charmbracelet/lipgloss"
)

func MultiSelect(choices []string) ([]string, error) {
	items := make([]list.Item, 0, len(choices))
	for _, c := range choices {
		items = append(items, item{title: c})
	}

	slices.SortFunc(items, func(a, b list.Item) int {
		return cmp.Compare(a.(item).title, b.(item).title)
	})

	l := list.New(items, itemDelegate{}, 0, 0)
	l.Title = "Select repos to register"
	l.AdditionalShortHelpKeys = extraKeys
	l.AdditionalFullHelpKeys = extraKeys

	height := 30 // 20 + 10, 10 is a magic number to show 20 items
	if size := len(items); size < 20 {
		height = size + 10
	}
	model := model{list: l, height: height}
	p := tea.NewProgram(model)
	if _, err := p.Run(); err != nil {
		return nil, err
	}

	selection := make([]string, 0, len(items))
	for _, listItem := range items {
		item := listItem.(item)
		if item.checked {
			selection = append(selection, item.title)
		}
	}

[...]

PR #574, opened 2024-07-31
feat(list): implement GlobalIndex helper

nobe4:
This commit introduces a new `filteredItem` field called `index` which stores the index of the filtered item in the unfiltered list. This allows to get at runtime the unfiltered list index for the selected (and possibly filtered) item.

This is the only solution I found to use `SetItem` with a filtered list.

The name `GlobalIndex` might not be ideal, I'm happy to change it to something else. (`UnfilteredIndex`?)

Fixes: #550

<details><summary>Example</summary>

Consider the following code:

```go
package main

import (
	"fmt"
	"io"
	"log"
	"os"

	"github.com/charmbracelet/bubbles/list"
)

type item struct {
	title   string
	checked bool
}

func (i item) Title() string       { return i.title }
func (_ item) Description() string { return "" }
func (i item) FilterValue() string { return i.title }

type itemDelegate struct{}

func (_ itemDelegate) Height() int                             { return 1 }
func (_ itemDelegate) Spacing() int                            { return 0 }
func (_ itemDelegate) Update(_ tea.Msg, _ *list.Model) tea.Cmd { return nil }
func (_ itemDelegate) Render(w io.Writer, m list.Model, index int, listItem list.Item) {
	i, ok := listItem.(item)
	if !ok {
		return
	}

	cursor := " "
	if index == m.Index() {
		cursor = ">"
	}

	checked := " "
	if i.checked {
		checked = "x"
	}

	fmt.Fprint(w, cursor+checked+i.title)
}

type model struct {
	list list.Model
}

	switch msg := msg.(type) {

	case tea.KeyMsg:
		switch msg.String() {
		case "q":
			return m, tea.Quit

		case " ":
			log.Print("using global index")
			i := m.list.SelectedItem().(item)
			i.checked = !i.checked
			return m, m.list.SetItem(m.list.GlobalIndex(), i)

		case "x":
			log.Print("using index")
			i := m.list.SelectedItem().(item)
			i.checked = !i.checked
			return m, m.list.SetItem(m.list.Index(), i)
		}
	}

	var cmd tea.Cmd
	m.list, cmd = m.list.Update(msg)
	return m, cmd
}

func (_ model) Init() tea.Cmd { return nil }
func (m model) View() string  { return m.list.View() }

func main() {
	choices := []string{"a", "b1", "b2", "b3", "c"}

	items := make([]list.Item, 0, len(choices))
	for _, i := range choices {
		items = append(items, item{title: i})
	}

	l := list.New(items, itemDelegate{}, 0, 0)
	l.SetHeight(20)

	f, err := tea.LogToFile("debug.log", "debug")
	if err != nil {
		fmt.Println("fatal:", err)
		os.Exit(1)
	}
	defer f.Close()

	model := model{list: l}
	p := tea.NewProgram(model)
		panic(err)
	}

	fmt.Printf("%#v\n", items)
}
```

Selecting with `" "` and `"x"` behaves similarly before filtering. Upon filtering, `"x"` behaves as described in #550 (i.e. wrongly) and `" "` behaves correctly: it updates the correct item in place.

</details>

### DIFF
--- a/list/list.go
+++ b/list/list.go
@@ -52,6 +52,7 @@ type ItemDelegate interface {
 }
 
 type filteredItem struct {
+	index   int   // index in the unfiltered list
 	item    Item  // item matched
 	matches []int // rune indices of matched items
 }
@@ -459,6 +460,18 @@ func (m Model) MatchesForItem(index int) []int {
 // entire slice of items.
 func (m Model) Index() int {
 	return m.Paginator.Page*m.Paginator.PerPage + m.cursor
+}
+
+// GlobalIndex returns the index of the currently selected item as it is stored
+// in the unfiltered list of items. This value can be used with SetItem().
+func (m Model) GlobalIndex() int {
+	index := m.Index()
+
+	if m.filteredItems == nil || index >= len(m.filteredItems) {
+		return index
+	}
+
+	return m.filteredItems[index].index
 }
 
 // Cursor returns the index of the cursor on the current page.
@@ -1227,6 +1240,7 @@ func filterItems(m Model) tea.Cmd {
 		filterMatches := []filteredItem{}
 		for _, r := range m.Filter(m.FilterInput.Value(), targets) {
 			filterMatches = append(filterMatches, filteredItem{
+				index:   r.Index,
 				item:    items[r.Index],
 				matches: r.MatchedIndexes,
 			})
--- a/list/list.go
+++ b/list/list.go
@@ -456,8 +456,10 @@ func (m Model) MatchesForItem(index int) []int {
 	return m.filteredItems[index].matches
 }
 
-// Index returns the index of the currently selected item as it appears in the
-// entire slice of items.
+// Index returns the index of the currently selected item as it is stored in the
+// filtered list of items.
+// Using this value with SetItem() might be incorrect, consider using
+// GlobalIndex() instead.
 func (m Model) Index() int {
 	return m.Paginator.Page*m.Paginator.PerPage + m.cursor
 }


caarlos0 (approved):

bashbunni:
Ah this is such a good catch! Thank you so much for working on this. I'm thinking it might make sense to integrate your change directly into `Index` rather than creating a new function. It seems that `Index` and `Cursor` tend to be the same value while in reality we should be preserving the item's original index as you've done here.

I just wrote a test that showcases the behaviour you've added with this PR + the `Index`/`Cursor` dynamic https://github.com/charmbracelet/bubbles/blob/testindex/list/list_test.go#L76-L142

Let me know if you're cool with me adding this test to your PR. I'll also double check with the original author to see their thoughts on `Index`'s current behaviour

bashbunni:
I'll want to refactor the test before merging so there are more conditions to satisfy. Right now it's more of a walkthrough on the behaviour of this change. If you have any suggestions to improve it as well, I'm all ears :)

nobe4:
Thanks for the review @bashbunni ❤️ 

> I'm thinking it might make sense to integrate your change directly into Index [...] we should be preserving the item's original index as you've done here.

I explicitly added a new function/field because my understanding of `Index` and `Cursor` differed from what I wanted:

- `Cursor`: The current item position in the visible list
- `Index`: The item position in the filtered list
- `GlobalIndex`: The item position in the unfiltered list

So if I want to get the visible position of the item, only `Cursor` can currently give me this information. And if I want the item position in the filtered list, only `Index` can give me this information.

E.g. (assuming the cursor is on the item: `GId` = `GlobalIndex`, `Id` = `Index`, `C` = `Cursor`)

```
Filter: A / Item per Page: 2 / Page: 1
GId  Id  C   Item  Comment
0    0   0   A     Hidden by page
1    x   x   B     Hidden by filter
2    1   1   A     Hidden by page
3    2   0   A     Visible
4    3   1   A     Visible
5    x   x   B     Hidden by filter
6    4   0   A     Hidden by page
7    x   x   B     Hidden by filter
8    5   1   A     Hidden by page
9    6   0   A     Hidden by page
```

It seems to me that the `Index` naming should be reversed to be more explicit and less confusing:

- `Index`: The position of the item in the list
- `FilteredIndex`: The position of the item in the filtered list

But that might require a bigger codechange and introduce a breaking change, so I understand if this is not ideal.

Alternatively, `Index` could become a struct to erase all confusion:

```go
type Index struct {
   Filtered    int
   Unfiltered  int
}

func (m Model) Index() int {
        filteredIndex := m.Paginator.Page*m.Paginator.PerPage + m.cursor
        unfilteredIndex := filteredIndex

	if m.filteredItems != nil && index < len(m.filteredItems) {
		unfilteredIndex = m.filteredItems[index].index
	}

        return Index {
            Filtered: filteredIndex,
            Unfiltered: unfilteredIndex,
        }
}
```

But this would be even more disruptive to existing codebase.

> Let me know if you're cool with me adding this test to your PR. If you can think of a case where you need Index to remain as-is where GlobalIndex wouldn't fit the bill, please let me know! + I'll want to refactor the test before merging so there are more conditions to satisfy.

Go ahead, and I'll have a look once you've done so :) Feel free to ping me!
Thanks for writing this test!

bashbunni:
Thank you!! I will take another look when I have a moment. I do think this is a valuable fix here 🙏

nobe4:
Hi @bashbunni 👋 
Any updates here? I have another issue coming up that would benefits from this as well 😬

Let me know if / how I can help push this forward 🙏

g026r:
Might I recommend updating the godoc for `Index()` while you're doing this?

It currently states the following, which implies that it operates across the unfiltered list:
```
// Index returns the index of the currently selected item as it appears in the
// entire slice of items.
```

list/list.go
```diff
@@ -455,12 +456,26 @@ func (m Model) MatchesForItem(index int) []int {
 	return m.filteredItems[index].matches
 }
 
-// Index returns the index of the currently selected item as it appears in the
-// entire slice of items.
+// Index returns the index of the currently selected item as it is stored in the
+// filtered list of items.
+// Using this value with SetItem() might be incorrect, consider using
+// GlobalIndex() instead.
```
nobe4, line 462: @g026r done in [11d4e11](https://github.com/charmbracelet/bubbles/pull/574/commits/11d4e111405f1b7677766270f6de79311e43bf7c)

caarlos0 (approved):

bashbunni (approved):

bashbunni:
@nobe4 thank you so much for bringing this to our attention and for your patience!

PR #584, opened 2024-08-14
feat(key): accept any stringer interface to match keys

aymanbagabas:
This allows any `fmt.Stringer` interface to match key bindings. We need this in the upcoming v2 changes.

### DIFF
--- a/key/key.go
+++ b/key/key.go
@@ -127,7 +127,7 @@ type Help struct {
 }
 
 // Matches checks if the given KeyMsg matches the given bindings.
-func Matches[T fmt.Stringer](k T, b ...Binding) bool {
+func Matches[Key fmt.Stringer](k Key, b ...Binding) bool {
 	keys := k.String()
 	for _, binding := range b {
 		for _, v := range binding.keys {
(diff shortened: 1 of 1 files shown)


caarlos0 (approved):

key/key.go
```diff
@@ -129,7 +127,7 @@ type Help struct {
 }
 
 // Matches checks if the given KeyMsg matches the given bindings.
-func Matches(k tea.KeyMsg, b ...Binding) bool {
+func Matches[T fmt.Stringer](k T, b ...Binding) bool {
```
meowgorithm, line 130:
```suggestion
func Matches[Key fmt.Stringer](k Key, b ...Binding) bool {
```

key/key.go
```diff
@@ -129,7 +127,7 @@ type Help struct {
 }
 
 // Matches checks if the given KeyMsg matches the given bindings.
```
meowgorithm, line 129:
```suggestion
// Matches checks if the given key matches the given bindings.
```

meowgorithm (commented):
Small change to make working with LSPs more intuitive

caarlos0 (approved):

Issue #585, opened 2024-08-15
table's `StyleFunc` strips default cell padding


bashbunni:
When `StyleFunc` is set, the alignment of the header and body is off. 

**To Reproduce**
I created a failing test that reproduces the issue along with an example you can run to help debug this issue.
https://github.com/charmbracelet/bubbles/blob/stylefunc/table/table_test.go#L144

It looks like the body of the table is stripping left padding from the cells when `StyleFunc` is used

![image](https://github.com/user-attachments/assets/f602306f-8bea-41a8-9db2-dabb63ee4da4)

PR #504, opened 2024-04-11
Table add WithStyleFunc to customize cell styling 

KevM:
This change ports a [customization from lipgloss's table](https://github.com/charmbracelet/lipgloss/blob/29fafad5be923707318d4557fb39e80ffc0fa6ae/table/table.go#L10) to bubbles' table. 

## Option WithStyleFunc

This option lets the user set a style selection func during table creation. 

```go
type TableStyleFunc func(row, col int, value string) lipgloss.Style
```
This func allows the user to customize the style that will appear for each row and col. 

## Example 

It is tricky to commit a demo due to the examples for bubbles being in the bubbletea repository but this diff to the table example creates the result below where the Mexican row is customized to include the flag colors of Mexico.

### Example Diff 

```diff
diff --git a/examples/table/main.go b/examples/table/main.go
index 76736f3..23443ee 100644
--- a/examples/table/main.go
+++ b/examples/table/main.go
@@ -157,13 +157,6 @@ func main() {
                {"100", "Montreal", "Canada", "4,276,526"},
        }
 
-       t := table.New(
-               table.WithColumns(columns),
-               table.WithRows(rows),
-               table.WithFocused(true),
-               table.WithHeight(7),
-       )
-
        s := table.DefaultStyles()
        s.Header = s.Header.
                BorderStyle(lipgloss.NormalBorder()).
@@ -174,7 +167,29 @@ func main() {
                Foreground(lipgloss.Color("229")).
                Background(lipgloss.Color("57")).
                Bold(false)
-       t.SetStyles(s)
+
+       t := table.New(
+               table.WithColumns(columns),
+               table.WithRows(rows),
+               table.WithFocused(true),
+               table.WithHeight(7),
+               table.WithStyles(s),
+               table.WithStyleFunc(func(row, col int, value string) lipgloss.Style {
+                       if row == 5 {
+                               if col == 1 {
+                                       return s.Cell.Copy().Background(lipgloss.Color("#006341"))
+                               } else if col == 2 {
+                                       return s.Cell.Copy().Background(lipgloss.Color("#FFFFFF"))
+                               } else if col == 3 {
+                                       return s.Cell.Copy().Background(lipgloss.Color("#C8102E"))
+                               } else {
+                                       return s.Cell
+                               }
+                       }
+
+                       return s.Cell
+               }),
+       )
```
image
 

## Why 

[...]

PR #586, opened 2024-08-16
remove `StyleFunc` + add tests for table alignment

bashbunni:
Reverts https://github.com/charmbracelet/bubbles/pull/504 which was causing some alignment issues in the table. I do still think we should have this feature, but I'll save it for a future release when we replace table's rendering with `lipgloss.Table`

Closes https://github.com/charmbracelet/bubbles/issues/585

+ Added tests for table alignment to prevent this issue in future changes to the bubbles table


bashbunni:
Not sure why `lint` is failing here but fine on `master` 🤔

tyliec:
👍 on re-adding the functionality https://github.com/charmbracelet/bubbles/pull/504 added... eventually 😿

PR #589, opened 2024-08-19
feat(progress): model shouldn't conform to tea.Model

aymanbagabas:
The `progress` bubble is the only bubble that conforms to `tea.Model`. Remove `Init()` and make it return its own `Model`.


caarlos0:
I'm in favor of this change, although its a breaking change, so we should prob release this on v2.

It might be good to document somewhere that the bubbles aren't supposed to be `tea.Model`s, to avoid confusion about this in the future...

caarlos0 (approved):

PR #590, opened 2024-08-20
feat: add HelpView to table as a convenience

meowgorithm:


bashbunni:
Nice, this is a great addition

meowgorithm:
So this works great but lint is freaking out a lil' 🤔

PR #608, opened 2024-09-02
fix(viewport): preempt potential panics when calculating visible lines

meowgorithm:
This adds some checks to prevent slice out-of-bounds errors when calculating visible lines in the viewport.

This will fix the panic we're seeing in https://github.com/charmbracelet/gum/issues/660#issuecomment-2323328662 (though we still need to address the zero-height-by-default thing).

### DIFF
--- a/viewport/viewport.go
+++ b/viewport/viewport.go
@@ -117,12 +117,17 @@ func (m Model) maxYOffset() int {
 // visibleLines returns the lines that should currently be visible in the
 // viewport.
 func (m Model) visibleLines() (lines []string) {
-	if len(m.lines) > 0 {
-		top := max(0, m.YOffset)
-		bottom := clamp(m.YOffset+m.Height, top, len(m.lines))
-		lines = m.lines[top:bottom]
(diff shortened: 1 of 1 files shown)


viewport/viewport.go
```diff
@@ -117,12 +117,17 @@ func (m Model) maxYOffset() int {
...
-	if len(m.lines) > 0 {
-		top := max(0, m.YOffset)
-		bottom := clamp(m.YOffset+m.Height, top, len(m.lines))
-		lines = m.lines[top:bottom]
+	if len(m.lines) == 0 {
+		return nil
+	}
+	top := max(0, m.YOffset)
+	bottom := min(m.YOffset+m.Height, len(m.lines))
+	if top >= bottom {
```
caarlos0, line 125: do we also need to check if bottom is within m.lines bounds?

caarlos0:
not sure if this is still needed

meowgorithm:
Superseded and implemented. Closing.

Issue #9, opened 2020-10-20
Add plus/minus button widget please


mashnode:
Just like this:
![image](https://user-images.githubusercontent.com/73186958/96620102-904d9800-1339-11eb-8663-351baaf0419c.png)

PR #621, opened 2024-09-20
feat: add picker component

Broderick-Westrope:
Resolves https://github.com/charmbracelet/bubbles/issues/9.

Adds a new "picker" component. By default this component takes an implementation of the `State` interface and displays it in a horizontal picker manner (as requested in the issue).

- Two implementations of the `State` interface are included:
  - `ListState` takes a generic slice/list to display.
  - `IntState` allows you to set a maximum and minimum integer, or decide to ignore max or min.
- In addition to moving incrementally through the choices you can also enable "jumping" which will go to the first/last element (or max/min in the case of `IntState`).
- "Cycling" is another option which allows you to go past the first/last item (or max/min) and it will loop back to the other end of the list.
  - For `IntState` you can enable this on a per-end basis. Meaning you can set a min of 0 and max of 5, and also ignore the max. Going above 5 will continue incrementing, but going below 0 will set the value to 5. This can of course be enabled or disable for both min and max.
- By default indicators will be displayed to show when the user can increment / decrement the value.
- Methods are also included on the `Model` to get the individual, styled display components (indicators and the value). This allows the user to build a visually different component (such as a vertical picker).

### DIFF
--- a/picker/intstate.go
+++ b/picker/intstate.go
@@ -3,18 +3,25 @@ package picker
 type IntState struct {
 	min       int
 	max       int
+	selection int
 	ignoreMin bool
 	ignoreMax bool
-	selection int
 }
 
-func NewIntState(min, max int, ignoreMin, ignoreMax bool) *IntState {
+func NewIntState(min, max, selection int, ignoreMin, ignoreMax bool) *IntState {
+	switch {
+	case selection < min && !ignoreMin:
+		selection = min
+	case selection > max && !ignoreMax:
+		selection = max
+	}
+
 	return &IntState{
 		min:       min,
 		max:       max,
 		ignoreMin: ignoreMin,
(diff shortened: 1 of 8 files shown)


meowgorithm:
This’ll be a fun one. A few quick thoughts on very first read:

* `NewModel` should be `New`
* I’d add an ‘increment’ setting so you can change the amount you can increment by
* The shift modifier should allow you to increment by increment * 10
* Should the unit be a float so you can increment by, say 0.1?

Broderick-Westrope:
@meowgorithm thanks for the quick feedback!

These commits fix the `New` method and add "stepping" (incrementing by a set count using the shift modifier; uses 10 by default). Keen to hear more thoughts when you have a chance :)

P.S. Flexible on the naming of stepping, jumping, etc. I just went with the first things that came to mind.

PR #622, opened 2024-09-22
feat: Substitute mutex with atomic operations.

weezel:
Use atomic operations instead a mutex. This approach avoids locking and increment can be done within a single clock cycle, hence no deadlock nor race condition can occur.

### DIFF
--- a/filepicker/filepicker.go
+++ b/filepicker/filepicker.go
@@ -7,7 +7,7 @@ import (
 	"sort"
 	"strconv"
 	"strings"
-	"sync"
+	"sync/atomic"
 
 	"github.com/charmbracelet/bubbles/key"
 	tea "github.com/charmbracelet/bubbletea"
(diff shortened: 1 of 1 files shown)


weezel:
A gentle one week bump

weezel:
A gentle two weeks bump

bashbunni (approved):
Looks good to me! Will get another team member to also confirm before merging :)

meowgorithm:
Thanks for your patience with this one @weezel. If we think this makes sense we should probably do this across the board in Bubbles where we employ this pattern (for example, in `spinner`).

Before we merge this we should also look into why this is failing in CI.

filepicker/filepicker.go
```diff
@@ -7,25 +7,20 @@ import (
...
+var lastID atomic.Int64
 
 // Return the next ID we should use on the Model.
 func nextID() int {
-	idMtx.Lock()
-	defer idMtx.Unlock()
-	lastID++
-	return lastID
+	lastID.Add(1)
+	return int(lastID.Load())
```
caarlos0, line 23: add returns the new value, so this can be:

```suggestion
	return int(lastID.Add(1))
```

caarlos0:
> Before we merge this we should also look into why this is failing in CI.

`atomic.Int64` is not available in go 1.18... should we bump it?

meowgorithm:
@weezel, just a heads up that we merged this in #634 based on your implementation. Thanks for flagging this one, and for the fix ❤️

weezel:
Thanking you 😊👍

Issue #606, opened 2024-08-30
List's `setSize` method should also call `updateKeybindings`


michaelriri:
**Describe the bug**

Keybindings are not updated (enabled/disabled) after resizing the  height of a list. 

**To Reproduce**

You can see this behavior by calling `SetHeight` in the [list-simple](https://github.com/charmbracelet/bubbletea/tree/main/examples/list-simple) example. 

```go
func (m model) Update(msg tea.Msg) (tea.Model, tea.Cmd) {
	switch msg := msg.(type) {
	case tea.WindowSizeMsg:
		m.list.SetWidth(msg.Width)
		m.list.SetHeight(msg.Height - 1)
...
```

**Expected behavior**

The keybindings should update according to the List's height. For example, if there is only one page, then disable `next page` and `prev page` bindings. 

**Screenshots**

![image](https://github.com/user-attachments/assets/de596747-637b-43a3-8d50-54d3fd30fbca)
> [!NOTE]
> The `next page` & `prev page` keybindings are still enabled, even though there is only one page. 

**Additional context**

As a workaround, you can force the keybindings to update by calling `SetFilteringEnabled` which will, in turn, call `updateKeybindings`, but this is hacky. It seems like the straightforward fix is to also call `updateKeybindings` in `setSize`.

activey:
When this will be merged??

meowgorithm:
This is done and will be available in the next release. Appreciate your patience on this one, everyone.

PR #623, opened 2024-09-22
fix(list): update list keybindings after altering display size

Broderick-Westrope:
Resolves https://github.com/charmbracelet/bubbles/issues/606.

Updates the keybindings of the `list` component on each call to the `setSize` method. This resolves issues where pagination bindings are enabled (and shown in help) despite there now being only one page.


meowgorithm (approved):
LGTM. Appreciate the fix, and your patience with this one.

PR #624, opened 2024-09-23
feat(cursor): Add a method to retrieve the character underneath the cursor.

kyleochata:
Added a method to get the character beneath the Cursor. 

- `func (m Model) GetChar() string`

Ran into an edge case that can benefit from just knowing the character that is underneath the cursor.

### DIFF
--- a/cursor/cursor.go
+++ b/cursor/cursor.go
@@ -209,6 +209,11 @@ func (m *Model) SetChar(char string) {
 	m.char = char
 }
 
+// GetChar retrieves the character under the cursor.
+func (m Model) GetChar() string {
+	return m.char
+}
+
 // View displays the cursor.
 func (m Model) View() string {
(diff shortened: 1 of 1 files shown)


meowgorithm:
Interesting? What's the edge case?

kyleochata:
After getting some sleep I realized I overlooked `textinput.Position()` while navigating through a `textinput`.

Could be useful as a quality of life change for a character check rather than calling `Position()` and `Value()`

meowgorithm:
Would love to hear your actual use case so as to better ascertain if something like this makes sense.

bashbunni:
Hey @kyleochata thanks for looking for ways to improve bubbles! I've just created a discussion in place of this PR to see if others are interested in this feature. We can continue any discussion over there :) https://github.com/charmbracelet/bubbles/discussions/710

PR #625, opened 2024-09-23
fix: split line based on width of viewport

Ahmed-hassan99:
Fix content rendering in viewport by splitting each line based on viewport width. 

Previous implementation had a bug where lines exceeding the width would show in the viewport on multiple lines but the number of lines `len(m.lines)` would not reflect that. This results in the viewport rendering less lines than intended since line count is purely based on `\n` regardless of how many lines the content actually occupies on the terminal.

### DIFF
--- a/viewport/viewport.go
+++ b/viewport/viewport.go
@@ -102,7 +102,25 @@ func (m Model) ScrollPercent() float64 {
 // SetContent set the pager's text content.
 func (m *Model) SetContent(s string) {
 	s = strings.ReplaceAll(s, "\r\n", "\n") // normalize line endings
-	m.lines = strings.Split(s, "\n")
+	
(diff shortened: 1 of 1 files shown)


Ahmed-hassan99:
@meowgorithm 
Let me know if anything needs to change for PR description/ commit message.

PS It's my first open source PR so not sure if there's any standards/guidelines I missed

caarlos0:
I believe this should be solved by #697

bashbunni:
As Carlos stated above, this is fixed in his PR #697 and will be shipped as part of `v2`. If you feel otherwise, please don't hesitate to let us know. In the meantime I'm going to close this PR as resolved. Thank you so much for your support and efforts to make `viewport` better!

PR #629, opened 2024-09-30
feat(filepicker): add file selection persistence

CoMfUcIoS:
- Introduced `FileSelected` field to `Model` to store the selected file name.
- Updated `Update` method to restore the selected file index based on `FileSelected`.
- Adjusted min and max values to ensure the selected file is within the visible range.

Ex.
``` 
fp := filepicker.New()
fp.AllowedTypes = []string{".mod", ".sum", ".go", ".txt", ".md"}
fp.CurrentDirectory, _ = os.Getwd()
fp.FileSelected = "main.go"
```

### DIFF
--- a/filepicker/filepicker.go
+++ b/filepicker/filepicker.go
@@ -50,6 +50,7 @@ func New() Model {
 		maxStack:         newStack(),
 		KeyMap:           DefaultKeyMap(),
 		Styles:           DefaultStyles(),
+		FileSelected:     "",
 	}
 }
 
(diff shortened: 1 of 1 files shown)


bashbunni:
Hey @CoMfUcIoS you can do this with the current API, though we manage the persisted state through the model that owns the filepicker. Is there any reason the implementation in this PR might be preferable to the existing way of doing this?

Here's a working example of what I'm talking about: https://github.com/charmbracelet/bubbletea/blob/main/examples/file-picker/main.go

CoMfUcIoS:
At the point of raising this PR there wasnt a way to do it (as far as i remember :)). If it is now there then this makes this PR redundant, hence closing.

PR #630, opened 2024-10-02
feat(textinput): expose `UpdateSuggestions` method

nobe4:
This enables clients of the lib to update the suggestions at runtime whenever they please.

One direct application of this is when using `m.SetValue()`: After calling it, there is no way to get the updated `m.CurrentSuggestion()` as `m.SetValue()` does not update the suggestion list.

It would be possible to also run `m.updateSuggestions()` somewhere in `m.SetValue`, but that might interfere with other things.

> [!NOTE]
> Doing the following would work, but would require some useless re-calculation.
>  ```go
>  m.SetValue("...")
>  m.SetSuggestions(m.AvailableSuggestions())
>  m.CurrentSuggestion()
>  ```
> 
> See https://github.com/nobe4/gh-not/pull/188/files#diff-9a8a064fb3b750f10c2ed5e1dde89d98645d2984d15b8b541501b5f6829e32e1R25-R30 for a real-life usage.

### DIFF
--- a/textinput/textinput.go
+++ b/textinput/textinput.go
@@ -262,7 +262,7 @@ func (m *Model) SetSuggestions(suggestions []string) {
 		m.suggestions[i] = []rune(s)
 	}
 
-	m.updateSuggestions()
+	m.UpdateSuggestions()
 }
 
 // rsan initializes or retrieves the rune sanitizer.
@@ -621,7 +621,7 @@ func (m Model) Update(msg tea.Msg) (Model, tea.Cmd) {
 
 		// Check again if can be completed
 		// because value might be something that does not match the completion prefix
-		m.updateSuggestions()
+		m.UpdateSuggestions()
 
 	case pasteMsg:
 		m.insertRunesFromUserInput([]rune(msg))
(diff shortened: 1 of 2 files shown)


meowgorithm:
I think this make sense, though I do wonder: are they any use cases where we would not want this to be automatically called in `m.SetValue()` like you suggest?

nobe4:
I don't really know if there are usecases against automatically calling it in `m.SetValue()`.

Typically, the suggestions should be updated when:
- the suggestions changes ✅ [ref](https://github.com/nobe4/bubbles/blob/4382fdf1c6cfe18d34615c1e8b94b068022a461a/textinput/textinput.go#L258-L266)
- the value change via keypress ✅ [ref](https://github.com/nobe4/bubbles/blob/4382fdf1c6cfe18d34615c1e8b94b068022a461a/textinput/textinput.go#L622-L624)
- the value change via `m.SetValue()` ❌ 

So maybe just calling it in `m.SetValue()` would bring it closer to the expected behavior?

In any case I'm happy for either solution :)

PR #631, opened 2024-10-03
feat(textinput): store and recover history

nobe4:
This adds the history functionality to the textinput model. Specific actions can push a new value into the history and they can be switched back to via key presses.

This is done in relation to https://github.com/nobe4/gh-not/issues/190 I was considering implementing that in my project, and realized that it could just be part of bubbles instead.

Looking forward to getting some feedback on this. Especially the comments below.

<details><summary>Example usage</summary>

```go
package main

import (
)

type Model struct {
	t textinput.Model
}

func (m Model) Init() tea.Cmd { return nil }
func (m Model) Update(msg tea.Msg) (tea.Model, tea.Cmd) {
	switch msg := msg.(type) {
	case tea.KeyMsg:
		switch msg.Type {
		case tea.KeyEsc:
			return m, tea.Quit

		case tea.KeyEnter:
			m.t.SaveHistory()
			return m, nil
		}
	}

	var cmd tea.Cmd
	m.t, cmd = m.t.Update(msg)
	return m, cmd
}

	return m.t.View()
}

func main() {
	m := Model{
		t: textinput.New(),
	}
	m.t.Focus()

	p := tea.NewProgram(m)
		panic(err)
	}
}
```

</details>

### DIFF
--- a/textinput/textinput.go
+++ b/textinput/textinput.go
@@ -59,6 +59,8 @@ type KeyMap struct {
 	AcceptSuggestion        key.Binding
 	NextSuggestion          key.Binding
 	PrevSuggestion          key.Binding
+	NextHistory             key.Binding
+	PrevHistory             key.Binding
 }
 
 // DefaultKeyMap is the default set of key bindings for navigating and acting
@@ -78,8 +80,11 @@ var DefaultKeyMap = KeyMap{
 	LineEnd:                 key.NewBinding(key.WithKeys("end", "ctrl+e")),
 	Paste:                   key.NewBinding(key.WithKeys("ctrl+v")),
 	AcceptSuggestion:        key.NewBinding(key.WithKeys("tab")),
-	NextSuggestion:          key.NewBinding(key.WithKeys("down", "ctrl+n")),
-	PrevSuggestion:          key.NewBinding(key.WithKeys("up", "ctrl+p")),
+	NextSuggestion:          key.NewBinding(key.WithKeys("ctrl+n")),
+	PrevSuggestion:          key.NewBinding(key.WithKeys("ctrl+p")),
+	// TODO: discuss moving `down` and `up` to this instead of Next/PrevSuggestion
+	NextHistory: key.NewBinding(key.WithKeys("down")),
+	PrevHistory: key.NewBinding(key.WithKeys("up")),
 }
 
 // Model is the Bubble Tea model for this text input element.
@@ -152,6 +157,10 @@ type Model struct {
 	suggestions            [][]rune
 	matchedSuggestions     [][]rune
 	currentSuggestionIndex int
+
+	// history of inputs
+	history      [][]rune
+	historyIndex int
 }
 
 // New creates a new model with default settings.
@@ -170,6 +179,9 @@ func New() Model {
 		value:       nil,
 		focus:       false,
 		pos:         0,
+
+		history:      [][]rune{},
+		historyIndex: 0,
 	}
 }
 
@@ -184,6 +196,7 @@ func (m *Model) SetValue(s string) {
 	// caller. This avoids bugs due to e.g. tab characters and whatnot.
 	runes := m.san().Sanitize([]rune(s))
 	err := m.validate(runes)
+	m.SaveHistory()
 	m.setValueInternal(runes, err)
 }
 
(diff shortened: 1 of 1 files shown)


textinput/textinput.go
```diff
@@ -899,3 +919,56 @@ func (m Model) validate(v []rune) error {
...
+	}
+
+	// XXX: It seems that doing `append(m.history, m.value)` copies a reference
+	// to the value, which leads to all history items being similar.
+	// This is quite surprising given this other test:
+	// https://go.dev/play/p/MZGpcUzSVch
+	newRecord := make([]rune, len(m.value))
+	for i, r := range m.value {
+		newRecord[i] = r
+	}
```
nobe4, line 939: This feels suboptimal, but I didn't find any alternative solution.

textinput/textinput.go
```diff
@@ -78,8 +80,11 @@ var DefaultKeyMap = KeyMap{
 	LineEnd:                 key.NewBinding(key.WithKeys("end", "ctrl+e")),
 	Paste:                   key.NewBinding(key.WithKeys("ctrl+v")),
 	AcceptSuggestion:        key.NewBinding(key.WithKeys("tab")),
-	NextSuggestion:          key.NewBinding(key.WithKeys("down", "ctrl+n")),
-	PrevSuggestion:          key.NewBinding(key.WithKeys("up", "ctrl+p")),
+	NextSuggestion:          key.NewBinding(key.WithKeys("ctrl+n")),
+	PrevSuggestion:          key.NewBinding(key.WithKeys("ctrl+p")),
+	// TODO: discuss moving `down` and `up` to this instead of Next/PrevSuggestion
+	NextHistory: key.NewBinding(key.WithKeys("down")),
+	PrevHistory: key.NewBinding(key.WithKeys("up")),
```
nobe4, line 87: Up and Down feel more natural to me, as a shell/vim user than updating completion. However this might break existing applications and change the default behavior, so I'd be happy not to change it.

Which keys would you recommend?

textinput/textinput.go
```diff
@@ -899,3 +919,56 @@ func (m Model) validate(v []rune) error {
...
+		newRecord[i] = r
+	}
+
+	m.history = append(m.history, newRecord)
+	m.historyIndex = len(m.history) - 1
+}
+
+func (m Model) History() []string {
+	// TODO: rename getSuggestions to getStrings
+	return m.getSuggestions(m.history)
```
nobe4, line 947: The `m.getSuggestions` really just transforms a `[][]rune` into a `[]string`, so the name could be updated to reflect this.

meowgorithm:
Thanks for the PR! So right off the cuff, I sort of wonder if this should be implemented externally, _possibly_ with an interface. That would allow to, for example, store large amounts of history via sqlite, or store and fetch history over the network.

caarlos0:
text input already have the suggestions feature, which can be used as a history.

fwiw we also have the same feature [in the works for textarea](https://github.com/charmbracelet/bubbles/compare/master...area).

if you don't like how suggestions work, you can also disable the keybinding and wrap textinput, setting the actual contents of the history as the value of the textinput (instead of ghost text).

nobe4:
> I sort of wonder if this should be implemented externally, possibly with an interface.

Possibly! Do you think this would be more worthwhile? If so, would the interface be a member of the `TextInput` model?

I like the idea of decoupling the two.

The interface could look like: 

```go
type HistoryManager interface {
  Save(string) error
  Get() (string, error)
  Next()
  Previous()
}
```

> text input already have the suggestions feature, which can be used as a history.
> if you don't like how suggestions work, [...]

It can, but if I want to use both suggestion and history, then I would be stuck :)
And they serve slightly different purpose as well, IMHO it's not good to conflate the two.

Coming from (Neo)Vim, having the completion + the history is incredibly useful to repeat command.

bashbunni:
Hey, there should be a way to do add your history to `textinput` suggestions. We'll get back to you with an example of this. That said, this feature seems out of scope to be handled by the bubbles library

bashbunni:
Hey @nobe4 here's an example grabbing bash history, as promised
https://gist.github.com/bashbunni/9e2c472f0fdd3181d06ce7fea97c09bd

nobe4:
@bashbunni thanks for the example!

The PR/request is not about the shell history, but about the internal bubble `input` history.
So your example doesn't really answer the question, and I disagree that this would be out of scope.

Can you have another look? Much appreciated :pray:

PR #634, opened 2024-10-07
fix: use atomic package for ids

caarlos0:
#622 won't work with go 1.18, this should.

based on PR by @weezel 

closes #622


bashbunni:
Niiiiice thank you!!

PR #639, opened 2024-10-15
feat: tree bubble

### LINKED ISSUE #233: Feature request: Tree model
Hello, I want to ask if the charmbracelet community would be interested in accepting a PR for adding a tree model to the project.

A while ago I needed one for a project of mine and at the time there was none available, so I created one.

Currently I think it reached a level of maturity that other people might be interested in using it.

If anyone can have a look at the [code](https://github.com/mariusor/bubbles-tree/blob/master/tree.go), and the [example](https://github.com/mariusor/bubbles-tree/blob/master/examples/tree/main.go), let me know if it would be accepted as a PR.

dlvhdr:
Introduces a tree bubble for TUIs that require a tree like model with expand/collapse/select functionality.

https://github.com/user-attachments/assets/b33820ef-a708-4526-939c-d50616698a07

## Intended Interactions

- All the viewport scrolling methods
- expand / collapse
- select
- later on - filter which will show the tree of matching items

## API

- Use `lipgloss.Tree` for the rendering
- Expose `SelectedNode()`
- Expose `SetSelectedNode()`
- Tree items
    - Have a YOffset that represents their vertical position in the tree
    - Will expose their `YOffset` in a getter so users can scroll to a node
	- Will be useful if put inside a viewport
- Expose a node's `depth` in a getter
	- so users can style a node based on its depth
	- should be done in lipgloss?

## Example Usage

See https://github.com/charmbracelet/bubbletea/pull/1190

Closes #233.

### DIFF
--- a/tree/tree.go
+++ b/tree/tree.go
@@ -216,12 +216,13 @@ func (m Model) Update(msg tea.Msg) (Model, tea.Cmd) {
 func (m Model) View() string {
 	treeView := m.viewport.View()
 
-	var help string
+	var sections []string
+	sections = append(sections, treeView)
 	if m.showHelp {
-		help = m.helpView()
+		sections = append(sections, m.helpView())
 	}
 
-	return lipgloss.JoinVertical(lipgloss.Left, treeView, help)
+	return lipgloss.JoinVertical(lipgloss.Left, sections...)
 }
 
 // SetScrollOff sets the minimal number of lines to keep visible above and below the selected node.
@@ -618,6 +619,16 @@ func setYOffsets(t *Node) {
 			above += child.Size() - 1
 		}
 	}
+}
+
+// ViewportYOffset returns the vertical offset of the tree's viewport.
+func (m *Model) ViewportYOffset() int {
+	return m.viewport.YOffset()
+}
+
+// SetViewportYOffset sets the vertical offset of the tree's viewport
+func (m *Model) SetViewportYOffset(yoffset int) {
+	m.viewport.SetYOffset(yoffset)
 }
 
 // YOffset returns the vertical offset of the selected node.


tree/tree.go
```diff
@@ -0,0 +1,723 @@
...
+		Background(lipgloss.Color("62")).
+		Foreground(lipgloss.Color("18")).
+		Bold(true)
+	s.nodeFunc = func(_ Nodes, _ int) lipgloss.Style {
+		return s.NodeStyle
+	}
+	s.selectedNodeFunc = func(_ Nodes, _ int) lipgloss.Style {
+		return s.SelectedNodeStyle
+	}
+	s.SelectionCursor = lipgloss.NewStyle().Foreground(lipgloss.Color("9"))
```
dlvhdr, line 46: currently this does nothing, but we could use it to show a pointing cursor. Is that something we want?

tree/tree.go
```diff
@@ -0,0 +1,723 @@
...
+	// Help toggle keybindings.
+	ShowFullHelp  key.Binding
+	CloseFullHelp key.Binding
+
+	Quit key.Binding
+}
+
+// DefaultKeyMap is the default set of key bindings for navigating and acting
+// upon the tree.
+var DefaultKeyMap = KeyMap{
```
dlvhdr, line 67: probably need to add more keybinds, I was think also about:
- half a page up / down which would equal 10 rows let's say?
- anything else?

tree/tree.go
```diff
@@ -0,0 +1,723 @@
...
+	height int
+	// yOffset is the vertical offset of the selected node.
+	yOffset int
+}
+
+// Node is a a node in the tree
+// Node implements lipgloss's tree.Node
+type Node struct {
+	// tree is used as the renderer layer
+	tree *ltree.Tree
```
dlvhdr, line 129: I'm thinking whether I should implement everything here, and not use `lipgloss.Tree` because:
1. I'm using lipgloss.Tree as a renderer, but I found myself implementing a lot of its logic.
2. I've also found that the `table` bubble doesn't actually use `lipgloss.Table`
3. I needed more granular styling like the notion of a selected node, which `lipgloss.Tree` doesn't "natively" support, so I wasn't sure if what I'm doing is an abuse.
4. I would also need to expose everything `lipgloss.Tree` exports like all the different Enumerator styles and the rest.

Let me know what you think. Should the view function also be implemented here and not use `lipgloss.Tree.View()`?

tree/tree.go
```diff
@@ -0,0 +1,723 @@
...
+// Size returns the number of nodes in the tree.
+// Note that if a child isn't open, its size is 1
+func (t *Node) Size() int {
+	return t.size
+}
+
+// YOffset returns the vertical offset of the Node
+func (t *Node) YOffset() int {
+	return t.yOffset
+}
```
dlvhdr, line 172: These can be used to allow for more flexible styling.
I found the `StyleFunc` of `lipgloss.Tree` doesn't support some stuff I want, like appending a symbol to the end of each item at the same indentation to the right of it. Maybe there are more use cases.

tree/tree.go
```diff
@@ -0,0 +1,723 @@
...
+
+type itemOptions struct {
+	openCharacter   string
+	closedCharacter string
+	treeYOffset     int
+}
+
+// Used to print the Node's tree
+// TODO: Value is not called on the root node, so we need to repeat the open/closed character
```
dlvhdr, line 181: Here I found I needed to implement the String() to repeat the open/close character because the `Value()` func isn't called on the root. Maybe this is an easy fix in lipgloss.

tree/tree.go
```diff
@@ -0,0 +1,723 @@
...
+		}
+		return s.Render(t.opts.closedCharacter + " " + t.tree.Value())
+	}
+	return s.Render(t.tree.Value())
+}
+
+// GivenValue returns the value passed to the node.
+func (t *Node) GivenValue() any {
+	return t.value
+}
```
dlvhdr, line 206: This was helpful in the TOC example I think. I found that the tree "swallows" my values and I can't read them again. This forces the user to save them before passing to the tree, which is just duplication and not the most ergonomic API. The name is confusing though given the `Value()` func also exists.

tree/tree.go
```diff
@@ -0,0 +1,723 @@
...
+
+// Children returns the children of an item.
+func (t *Node) Children() ltree.Children {
+	return t.tree.Children()
+}
```
dlvhdr, line 211: This exposes lipgloss, but it has to return `lipgloss.Children` since that's the interface

tree/tree.go
```diff
@@ -0,0 +1,723 @@
...
+	m.updateStyles()
+}
+
+// SetShowHelp shows or hides the help view.
+func (m *Model) SetShowHelp(v bool) {
+	m.showHelp = v
+}
+
+// SetSize sets the width and height of this component.
+func (m *Model) SetSize(width, height int) {
```
dlvhdr, line 489: Currently the `width` is applied only to the help model.
I saw other bubbles with width/height that also only apply them to select child models.
So I wasn't sure if:
1. Height should be applied anywhere
2. I need to wrap the `View()` return value in a style that applies them

tree/tree.go
meowgorithm: Ideally, we use `lipgloss/tree` for rendering. If it doesn't support what we need we can get it there.

So for some background, the `table/bubble` predates `lipgloss.Table`, but the plan was always to migrate to Lip Gloss for rendering as `lipgloss/table`, which is much, much more robust than the Bubbles implementation.

tree/tree.go
meowgorithm: Have you seen the [toggle example](https://github.com/charmbracelet/lipgloss/blob/master/examples/tree/toggle/main.go#L40-L53)? I actually put that one together in order to make sure `tree` can work interactively (though I of course may have missed some things).

tree/tree.go
meowgorithm: I think generally we can probably mirror viewport keybindings here.

Also, one thing we're doing in `v2` is removing globals when possible. So I'd change the above to a function:

```go
func DefaultKeyMap() KeyMap
```

tree/tree.go
```diff
@@ -0,0 +1,797 @@
...
+
+// View renders the component.
+func (m Model) View() string {
+	var leftDebugView, cursor string
+	// TODO: remove
+	if os.Getenv("DEBUG") == "true" {
+		leftDebugView = printDebugInfo(m.root) + " "
+		for i := 0; i < m.root.size; i++ {
+			if i == m.yOffset {
+				cursor = cursor + "👉 "
```
meowgorithm: We should make the 👉 configurable.

dlvhdr: Done

dlvhdr:
update: starting to take shape when used inside https://github.com/dlvhdr/diffnav

image

caarlos0 (commented):
this looks great! would be awesome to add an example as well

wjam:
I've just been playing with this PR locally and wondered whether it'd be possible to configure the open/closed character per node? The idea I had was a file tree where the closed character for a directory would be a closed folder and the open character would be an open folder.

dlvhdr:
hey guys :) any update on this?

bashbunni:
Hey @dlvhdr sorry for the delay on this. We're going to spend some energy cleaning up bubbles before v2, so rest assured this will be merged soon

vimkim:
This looks awesome! This will give better interactivity in its core essence. I think I'll make a lot of TUIs if this gets accepted. Thanks everyone for their efforts for making the SW world a better place!

tree/tree.go
```diff
@@ -0,0 +1,656 @@
...
+	// ScrollOff is the minimal number of lines to keep visible above and below the selected node.
+	ScrollOff int
+	// OpenCharacter is the character used to represent an open node.
+	OpenCharacter string
+	// ClosedCharacter is the character used to represent a closed node.
+	ClosedCharacter string
+	// CursorCharacter is the character used to represent the cursor.
+	CursorCharacter string
+	// KeyMap encodes the keybindings recognized by the widget.
+	KeyMap KeyMap
```
andreynering, line 117: I believe that in general, we'd like to avoid having public fields here as much as possible. So these keys like `OpenCharacter` and `ClosedCharacter` should be made private and additional `Set*` functions (`SetOpenCharacter`) should be implemented.

Am I right? @bashbunni @meowgorithm

tree/tree.go
```diff
@@ -0,0 +1,656 @@
...
+// func (m *Model) selectedNodeStyle() StyleFunc {
+// 	return func(children Nodes, i int) lipgloss.Style {
+// 		child := children.At(i)
+// 		if child.yOffset == m.yOffset {
+// 			return m.styles.selectedNodeFunc(children, i)
+// 		}
+//
+// 		return m.styles.nodeFunc(children, i)
+// 	}
+// }
```
andreynering, line 613: Commented code here. If not needed, delete.

tree/tree.go
```diff
@@ -0,0 +1,656 @@
...
+	}
+	return b
+}
+
+func max(a, b int) int {
+	if a > b {
+		return a
+	}
+	return b
+}
```
andreynering, line 656: Those are not needed anymore, as we recently upgraded the Go version.

`min` and `max` are not generic Go builtins.

andreynering (commented):
Hey @dlvhdr,

This looks pretty solid, overall!

Some comments:

- Can you rebase with `v2-exp` and squash into a minimal number of commits? (Can be a single commit)
- Is it easy enough to write some tests for this package?
- Ensure CI is green (lint included)
  - You might need to temporarily pin the last commit from https://github.com/charmbracelet/lipgloss/pull/446. Once merged, we'd pin from `v2-exp`
- I made a couple of comments below

tree/tree.go
```diff
@@ -0,0 +1,651 @@
...
+	openCharacter string
+	// closedCharacter is the character used to represent a closed node.
+	closedCharacter string
+	// cursorCharacter is the character used to represent the cursor.
+	cursorCharacter string
+	// KeyMap encodes the keybindings recognized by the widget.
+	KeyMap KeyMap
+	// styles sets the styling for the tree
+	styles Styles
+	Help   help.Model
```
andreynering, line 123: Can you move these two public fields (`KeyMap` and `Help`) to the top of the list, and add an empty line after to separate from the rest? The `table.Table` model also does this.

tree/tree.go
```diff
@@ -0,0 +1,651 @@
...
+
+	// Additional key mappings for the short and full help views. This allows
+	// you to add additional key mappings to the help menu without
+	// re-implementing the help component. Of course, you can also disable the
+	// list's help component and implement a new one if you need more
+	// flexibility.
+	AdditionalShortHelpKeys func() []key.Binding
+	AdditionalFullHelpKeys  func() []key.Binding
```
andreynering, line 131: I feel that these two should also be private with `Set*` functions.

go.mod
```diff
@@ -18,15 +21,24 @@ require (
...
+	github.com/muesli/ansi v0.0.0-20230316100256-276c6243b2f6 // indirect
 	github.com/muesli/cancelreader v0.2.2 // indirect
+	github.com/muesli/termenv v0.16.0 // indirect
 	github.com/xo/terminfo v0.0.0-20220910002029-abceb7e1c41e // indirect
 	golang.org/x/sync v0.12.0 // indirect
 	golang.org/x/sys v0.31.0 // indirect
+	golang.org/x/text v0.23.0 // indirect
 )
+
+replace github.com/charmbracelet/lipgloss/v2 v2.0.0-beta.1 => ../lipgloss
```
andreynering, line 44: OK, I got that it's not possible to pin from your branch (as I asked) because it lives in your private fork.

It's fine to keep as is for now, and I can manually pin to the branch commit once I merge https://github.com/charmbracelet/lipgloss/pull/446

tree/tree.go
dlvhdr: I've changed these, but just noted that these are public fields in the list component.
Let me know if you want me to revert / change the list as well.

andreynering: I still prefer to keep them private. It may have been by mistake that these got public on the other package.

andreynering (approved):
Awesome work! 👏 👏 👏 

Note that we're targeting v2.1.0 for this feature, so we're going to wait for v2.0.0 to be released before we merge your PRs.

dlvhdr:
Cool :) thanks for the reviews!

ossenthusiast:
@andreynering @dlvhdr any update on this?

ossenthusiast:
@dlvhdr can you please extract this in a separate repository, so people can use this without using forked bubbles?

andreynering:
@ossenthusiast As I said above, we're targeting v2.1.0 to ship this.

We're already on a Release Candidate for v2.0.0, so the final release is near! Once released, we can do a final review and merge this.

danvergara:
Please reopen the PR @dlvhdr 
Looking forward to having a native tree in bubbles

andreynering:
As I said above, we targeting v2.1 for this feature.

Now that v2.0 was released, this just need to be reopened targeting `main`.

dlvhdr:
Yup I'll rebase this 💪🏽

dlvhdr:
Reopened as https://github.com/charmbracelet/bubbles/pull/893

PR #643, opened 2024-10-21
feat: use inhouse wcwidth package instead of go-runewidth

aymanbagabas:
This change replaces the `go-runewidth` package with our in-house x/wcwidth package. This package is a simplified version of go-runewidth that uses the `golang.org/x/text` package for Unicode character width calculations.

We also can't be using both go-runewidth and uniseg to calculate character widths, so we've removed the uniseg package as well. Otherwise, we might end up with inconsistent character width in terminals that don't support grapheme clusters.


caarlos0 (approved):

caarlos0:
is this still relevant?

aymanbagabas:
> is this still relevant?

No, closing...

PR #646, opened 2024-10-23
(v2) tidiness and consistency changes

meowgorithm:
This brings some general consistency to Bubbles as well as general updates to make it work with Lip Gloss v2. A few things happening so far:

* Things like `DefaultStyles` and `DefaultKeymap` are becoming functions, not globals
* When applicable, focused and blurred styles are children of a larger style struct
* Sizing (widths, heights) are being standardized on via getters and setters
* Utility packages not intended for public consumption are being moved to `/internal`


meowgorithm:
Closing this to tackle things in smaller chunks.

PR #670, opened 2024-11-13
feat(textarea): set the real cursor position

aymanbagabas:
This makes the textarea bubble set the real cursor position on update. This need the new "ferocious" renderer to work.


caarlos0:
i wonder if it would make more sense to do this in the cursor pkg? 🤔 

then it could be used on both textarea, textinput, and others 🤔

caarlos0:
Yeah that makes sense 🤔

aymanbagabas:
> this is blocked by progress on ferocious right?

Correct

meowgorithm:
Closing this one; we're taking a new approach to real cursors in v2.

PR #673, opened 2024-11-20
List hide empty

mfulz:
Added a setting to list which can hide the "No items" on an empty list

### DIFF
--- a/list/list.go
+++ b/list/list.go
@@ -143,6 +143,7 @@ type Model struct {
 	showPagination   bool
 	showHelp         bool
 	filteringEnabled bool
+	hideEmpty        bool
 
 	itemNameSingular string
 	itemNamePlural   string
@@ -224,6 +225,7 @@ func New(items []Item, delegate ItemDelegate, width, height int) Model {
 		itemNameSingular:      "item",
 		itemNamePlural:        "items",
 		filteringEnabled:      true,
+		hideEmpty:             false,
 		KeyMap:                DefaultKeyMap(),
 		Filter:                DefaultFilter,
 		Styles:                styles,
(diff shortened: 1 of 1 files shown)


FFCoder:
I wonder if it would be better to just be called "HideEmptyStatus" rather than "SetHideEmpty"?

mfulz:
> I wonder if it would be better to just be called "HideEmptyStatus" rather than "SetHideEmpty"?

For me personally it doesn't matter, but I was following the other methods like: SetShowHelp, SetShowTitle, etc. which was the most obvious match to me. Especially if thinking of dynamic switching that one on and off, where HideEmptyStatus(false) would seem a bit weird to me :)

meowgorithm:
So part of me wonders if the API should be something like the following to provide more flexibility around how that empty items message is rendered.

```go
type Model structure {
    EmptyItemsNote func(plural string) string
}
```

And by default it would be something like:

```go
m.EmptyItemsNote = func(itemsPlural string) string {
    return fmt.Sprintf("(no %s)", itemsPlural)
}
```

And to hide it you'd:

```go
m.EmptyItemsNote = nil
m.RemoveEmptyItemsNote() // Maybe also a helper?
```

mfulz:
> So part of me wonders if the API should be something like the following to provide more flexibility around how that empty items message is rendered.
> 
> ```go
> type Model structure {
>     EmptyItemsNote func(string) string
> }
> ```
> 
> And by default it would be something like:
> 
> ```go
> m.EmptyItemsNote = func(itemsPlural string) string {
>     return fmt.Sprintf("(no %s)", itemsPlural)
> }
> ```
> 
> And to hide it you'd:
> 
> ```go
> m.EmptyItemsNote = nil
> m.RemoveEmptyItemsNote() // Maybe also a helper?
> ```

That's a way better idea :)

meowgorithm:
Okay cool! Do you mind updating accordingly?

mfulz:
> Okay cool! Do you mind updating accordingly?

Done :)

meowgorithm:
Thank you! This looks good on first glance: we'll give it a proper review shortly.

PR #680, opened 2024-11-26
feat: Granular progress bar completion

ChausseBenjamin:
A progress cell currently only has two states: complete, incomplete.

This PR let's progress bars be much more granular by having 9 possible states for the bar which drastically increases resolution.

Animations render at 60fps by default. It's a shame to have smoothness limited by progress bar resolution ;)

### DIFF
--- a/progress/progress.go
+++ b/progress/progress.go
@@ -3,6 +3,7 @@ package progress
 import (
 	"fmt"
 	"math"
+	"sort"
 	"strings"
 	"sync/atomic"
 	"time"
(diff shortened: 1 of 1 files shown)


ChausseBenjamin:
Don't mind the broken UI... This is just an example to show you guys how a better resolution can make animations smoother and make sliders much more precise/representative ;)

example

meowgorithm:
This is _SO_ good. We'll need to review this properly but I think its safe to say we should get this one in.

lrstanley:
> Also, the gradient implementation limits folks to 2 colors but a more generalized `gradientFunc` could allow people to do more funky stuff. Here's an example where I created a hue `gradientFunc` as well as a stretched gradient (where a and b from the gradient span from 0% to current % instead of the entire 100%) where the right-most completed color always remains the same.

Mentioned this in Discord, but referencing here for others as well: https://github.com/charmbracelet/bubbles/pull/838 -- that PR should work for what you're looking for, and should still be forward-compatible with some of the additional ideas you have. Though of course, the new implementation would be v2 only.

PR #682, opened 2024-11-27
Don't render suggestions when textinput is not focused.

elProxy:
This can look confusing in forms otherwise.

### DIFF
--- a/textinput/textinput.go
+++ b/textinput/textinput.go
@@ -665,7 +665,7 @@ func (m Model) View() string {
 		v += styleText(m.echoTransform(string(value[pos+1:]))) // text after cursor
 		v += m.completionView(0)                               // suggested completion
 	} else {
-		if m.canAcceptSuggestion() {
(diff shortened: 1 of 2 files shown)


bashbunni:
hey @elProxy could you please include a before and after for this change? It would make it easier to review with an example and/or visual

elProxy:
Hi @bashbunni 
Here's visually a quick before/after.
Before:
image
After:
image

I think it should actually be possible to write a simple test for it, so if you think the change makes sense in principle, I guess I could try and add one.

elProxy:
Hey there @bashbunni !
Sorry I didn't get around to that 2-liner earlier 😅 
Hope that helps.

elProxy:
@bashbunni I'd argue it's a simple bug fix... 😄

meowgorithm (approved):

meowgorithm:
Rebased on `master`. We're good to go. Thanks for the contribution, @elProxy!

PR #683, opened 2024-12-04
fix: allow to use empty style properly

caarlos0:
alternative to #673

This way users can set the text they want using `SetString` in the style.

closes #673

### DIFF
--- a/list/style.go
+++ b/list/style.go
@@ -67,7 +67,8 @@ func DefaultStyles() (s Styles) {
 		Foreground(lipgloss.AdaptiveColor{Light: "#A49FA5", Dark: "#777777"}).
 		Padding(0, 0, 1, 2) //nolint:mnd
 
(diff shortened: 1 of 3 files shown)


bashbunni (approved):
Nice

list/style.go
```diff
@@ -67,7 +67,8 @@ func DefaultStyles() (s Styles) {
 		Foreground(lipgloss.AdaptiveColor{Light: "#A49FA5", Dark: "#777777"}).
 		Padding(0, 0, 1, 2) //nolint:mnd
 
-	s.StatusEmpty = lipgloss.NewStyle().Foreground(subduedColor)
+	s.StatusEmpty = lipgloss.NewStyle().Foreground(subduedColor).
+		SetString("No items.")
```
aymanbagabas, line 71: Don't we need to do the same for `s.NoItems`?

caarlos0: done!

aymanbagabas (approved):

PR #717, opened 2025-01-27
(v2) update for generics-based design

meowgorithm:
This set of revisions makes the few necessary changes to for Bubbles to use the new generics-based design.


meowgorithm:
Closing as we're not going forward with this in Bubble Tea.

PR #719, opened 2025-01-29
refactor: progress: return Model instead of tea.Model

aymanbagabas:
This conforms to the new generic bubbletea API.
See https://github.com/charmbracelet/bubbletea/pull/1298


meowgorithm (approved):
Looks good. I also made this change in `v2-generics`.

PR #722, opened 2025-01-31
(v2) merge textarea API updates into v2-area

meowgorithm:
This brings textarea API updates (and other stuff on `v2-exp`) into `v2-area`.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,5 +1,4 @@
-Bubbles
-=======
+# Bubbles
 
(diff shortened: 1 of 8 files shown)


aymanbagabas (approved):

meowgorithm:
Tests are broken here, but the same ones are also broken in `v2-area` in the same way, so I'm going to merge.

PR #723, opened 2025-02-04
fix(viewport): handle nil LeftGutterFunc

aymanbagabas:
Since LeftGutterFunc is public, it can be nil if viewport is instantiated without using `viewport.New()`. This commit adds nil checks to handle this case.


andreynering (approved):

caarlos0:
ah, good catch

PR #727, opened 2025-02-06
feat: real cursor in textinput

caarlos0:
PS: we should probably use v2 as an opportunity to standarize some APIs, e.g.:

- Styles as a field the user can set
- Cursor wherever there are inputs
- ??


textinput/textinput.go
```diff
@@ -92,7 +94,19 @@ type Model struct {
...
+	VirtualCursor bool
+	cursor        cursor.Model
+
+	// ColumnOffset is the number of columns that the cursor is offset from the
+	// start of the line.
+	ColumnOffset int
+
+	// RowOffset is the number of rows that the cursor is offset from the start
+	// of the line.
+	RowOffset int
```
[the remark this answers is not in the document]
caarlos0: ok I think I gotcha what you mean, you're right, we can handle that in the caller!

caarlos0:
tested this on some bubbletea examples, seems to be fine 🙏🏻 

gonna merge it because i need it in the shell :)

textinput/textinput.go
```diff
@@ -92,16 +92,15 @@ type Model struct {
...
-	//
-	// For an introduction to styling with Lip Gloss see:
-	// https://github.com/charmbracelet/lipgloss
-	PromptStyle      lipgloss.Style
-	TextStyle        lipgloss.Style
-	PlaceholderStyle lipgloss.Style
-	CompletionStyle  lipgloss.Style
+	// VirtualCursor determines whether or not to use the virtual cursor. If
+	// set to false, use [Model.Cursor] to return a real cursor for rendering.
+	VirtualCursor bool
```
raphamorim, line 98: So textinput will start with VirtualCursor right?

raphamorim: sorry now I see it is the config to swich

Issue #744, opened 2025-02-26
List Width not respected


FFX01:
**Describe the bug**
When creating a new list, width is a required argument. This value does absolutely nothing unless it is set to a value smaller than the width of the widest list item. Looking through the code that implements list, I could not find anywhere in which the width member of the `list.Model` struct is used to force the rendering to the specified width. It does appear to be used in some capacity when rendering the filter input and the help section, but that's it. Even then, it does not appear to have an effect from my testing.

**Setup**
Please complete the following information along with version numbers, if applicable.
 - OS: KDE Neon
 - Shell: zsh
 - Terminal Emulator: Kitty
 - Terminal Multiplexer: N/A
 - Locale: US/UTF-8

**To Reproduce**
1. Use `list.New` to create a new list. Set the width to a value larger than the width of your widest list item.
2. render the list in your view function.
3. Observe that list does not respect tha value of the `width` argument.

**Source Code**
This is the source code from following the tutorial in this [video](https://www.youtube.com/watch?v=hF7qMEowPTw):
```go
package main

import (
	"log"

	"github.com/charmbracelet/bubbles/list"
	tea "github.com/charmbracelet/bubbletea"
	"github.com/charmbracelet/lipgloss"
)

type status int

const layoutDivisor int = 4

const (
	todo status = iota
	inProgress
	done
)

/* Styling */
var (
	columnStyle = lipgloss.NewStyle().
			Padding(2, 1).
			Border(lipgloss.HiddenBorder())
	focusedStyle = lipgloss.NewStyle().
			Padding(2, 1).
			Border(lipgloss.RoundedBorder()).
			BorderForeground(lipgloss.Color("62"))
	helpStyle = lipgloss.NewStyle().
			Foreground(lipgloss.Color("41"))
)

/* Task List Item */

type Task struct {
	title       string
	description string
	status      status
}

func (t *Task) Title() string       { return t.title }
func (t *Task) Description() string { return t.description }
func (t *Task) FilterValue() string { return t.title }

/* Main Model */
type Model struct {
	focused status
	lists   []list.Model
	err     error
	loaded  bool
}

func NewModel() *Model {
	m := new(Model)
	return m
}

func (m *Model) onNextList() {
	if m.focused == done {
		m.focused = todo
	} else {
		m.focused++
	}
}

func (m *Model) onPrevList() {
	if m.focused == todo {
		m.focused = done
	} else {
		m.focused--
	}
}

func (m *Model) initLists(width, height int) {
[...]

PR #749, opened 2025-03-15
Fix bug causing List.Width to not be respected

ian-howell:
This addresses a bug when rendering Lists that causes them to ignore the `width` value provided by `New` and `SetSize`. It works by extending the existing "truncate" logic with additional padding logic to ensure that each line of text is exactly the width of the list.

This addresses #744

### DIFF
--- a/list/defaultitem.go
+++ b/list/defaultitem.go
@@ -157,15 +157,16 @@ func (d DefaultDelegate) Render(w io.Writer, m Model, index int, item Item) {
 	}
 
 	// Prevent text from exceeding list width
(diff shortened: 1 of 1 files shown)


kalyncoose:
I believe the root cause is actually the help text rendering past the list's `m.width`, re: https://github.com/charmbracelet/bubbles/issues/744#issuecomment-2753712246

ian-howell:
The original issue didn't have the help message shown though...

kalyncoose:
> The original issue didn't have the help message shown though...

That's okay maybe both our PRs are needed then. Separate issues, same component.

ian-howell:
@meowgorithm Sorry to bother, but could I please get a review on this?

PR #755, opened 2025-03-25
ci: test on oldstable and stable go versions, automerge dependabot

caarlos0:
- stable and oldstable are the latest and previous versions, so we don't need to update this when a new release is out
- auto-merge dependabot PRs like we do on other projects

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -6,11 +6,9 @@ jobs:
   test:
     strategy:
       matrix:
-        go-version: ['1.23', '1.24']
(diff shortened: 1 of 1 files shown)


.github/workflows/build.yml
```diff
@@ -6,11 +6,9 @@ jobs:
   test:
     strategy:
       matrix:
-        go-version: ['1.23', '1.24']
+        go-version: [stable, oldstable]
```
andreynering, line 9: This is nice! I didn't know about it.

andreynering (approved):

PR #757, opened 2025-03-25
feat(viewport): expose XOffset field

tjbarber:
I've run into a situation where I want to be able to move the viewport horizontally by a variable amount, and that amount (for example) is dictated by the width off all the content visible and to the left of my screen subtracted by my screen's width plus the viewport's `XOffset`. This allows me to do something like this:

https://github.com/user-attachments/assets/6d9caee8-dd9f-414a-936f-418a783568d7

Considering that we expose the `YOffset` it seems to make sense that we should also expose the `XOffset` as well.

### DIFF
--- a/viewport/viewport.go
+++ b/viewport/viewport.go
@@ -39,8 +39,8 @@ type Model struct {
 	// YOffset is the vertical scroll position.
 	YOffset int
 
-	// xOffset is the horizontal scroll position.
-	xOffset int
+	// XOffset is the horizontal scroll position.
+	XOffset int
 
 	// horizontalStep is the number of columns we move left or right during a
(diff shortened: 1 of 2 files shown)


caarlos0:
#763 adds a `SetXOffset`

tjbarber:
@caarlos0 is there a reason why we can't expose this field? We're already exposing `YOffset`. It would be nice to be able to do something like `SetXOffset(m.viewport.XOffset + cell.Width)`

tjbarber:
@caarlos0 ah that's perfectly reasonable and fits my use case, thanks for the explanation. I'll go ahead and close this PR.

caarlos0:
awesome, thanks!

PR #759, opened 2025-03-26
filepicker: add support for abstracted filesystem io/fs.FS

dolmen:
Add an optional [`io/fs.FS`](https://pkg.go.dev/io/fs#FS) to `filepicker`'s model to allow to browse an abstracted filesystem instead of the full filesystem exposed by the [`os`](https://pkg.go.dev/os) package.

This allows for example to browse only a subdirectory by using [`os.Root`](https://pkg.go.dev/os#Root) (go1.24+):
```go
root, _ := os.OpenRoot(dir)                                                                           
fp.FS = root.FS()
```

Note that Go 1.25 will add [`io/fs.ReadLink`](https://pkg.go.dev/io/fs@master#ReadLink), but we can't use it yet. So we implement symlink resolution here.

By the way, references to [`os`](https://pkg.go.dev/os) types and constants which are aliases to same symbols in [`io/fs`](https://pkg.go.dev/io/fs) are replaced: `os.Mode`*, [`os.DirEntry`](https://pkg.go.dev/os#DirEntry).


dolmen:
Cc: @maaslalani (filepicker original author)

dolmen:
Closing because replaced by #954 upgraded to v2 API.

PR #766, opened 2025-03-28
Table: wrap cursor on line up and line down

joshallenit:
Wrap cursor on line up and line down.


joshallenit:
@aymanbagabas Hi, bump, looking for a review.

PR #768, opened 2025-03-31
fix(textinput): improve placeholder

caarlos0:
similar to #767 but for textinputs

### DIFF
--- a/.golangci.yml
+++ b/.golangci.yml
@@ -1,17 +1,6 @@
+version: "2"
 run:
   tests: false
-
-issues:
-  include:
-    - EXC0001
-    - EXC0005
-    - EXC0011
-    - EXC0012
-    - EXC0013
-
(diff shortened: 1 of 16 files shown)


andreynering (approved):
I could not reproduce by setting Chinese text [into this example](https://github.com/charmbracelet/bubbletea/blob/5c6d02d00b3f275d4a98b2dc475ca0c0ef928bf0/examples/textinput/main.go#L32). It already worked fine on `main` for me.

It keeps working after the patch, so still working.

bashbunni (approved):
Tests passing, new test case works well to highlight these changes. Looks great! :) BTW also tested the bubbletea textinput example locally with this change and it worked great for both arabic and japanese placeholders. The cursor blinking was also highlighting the entire japanese character which looks super clean (same for arabic).

image

PR #772, opened 2025-04-04
feat(table): v2 table bubble

bashbunni:
uses Lip Gloss table to render the table. Also allows users to create bubbles tables from existing Lip Gloss tables.

Depends on https://github.com/charmbracelet/lipgloss/pull/512


table/table.go
```diff
@@ -101,108 +92,370 @@ func DefaultKeyMap() KeyMap {
...
+		cursor:       0,
+		KeyMap:       DefaultKeyMap(),
+		Help:         help.New(),
+		table:        t,
+		useStyleFunc: true,
+	}
+	// We can't get the rows and headers from the table, so the user needs to
+	// provide them as arguments.
+	m.rows = rows
+	m.headers = headers
```
andreynering, line 135: FYI: I wrote a helper function on lipgloss to convert the internal `Data` into a `[][]string`:

https://github.com/charmbracelet/lipgloss/blob/c6b946e574dddb956f8642c9f84fa0219b777ed3/table/rows.go#L116

To avoid having to ask for the data again here, I think we should have a `GetHeaders()` and `GetData()` functions on lipgloss.

There is a small decision to make here. I'm not sure if `GetData()` should return a `[][]string` directly. For me, it makes more sense to make it return the internal `table.Data` interface and expose `dataToMatrix` as a public function.

table/table.go
```diff
@@ -101,108 +92,370 @@ func DefaultKeyMap() KeyMap {
...
+		m.SetHeaders(headers...)
 	}
 }
 
-// WithHeight sets the height of the table.
+// TODO andrey confirm this... I'm pretty sure that's how it's working now
+//
+// WithHeight sets the height of the table. The given height will be the total
+// table height including borders, margins, and padding. This function is used
+// as an [Option] in when creating a table with [New].
```
andreynering, line 258: I just made some quick tests, and it is still the full table size, and not the number of rows. It includes borders and white spaces.

I found a bug in the process, though. If less then 5, lipgloss is still allow the table to take 5 lines. I added a new topic on https://github.com/charmbracelet/glamour/issues/416 about this to fix soon.

table/table.go
```diff
@@ -101,108 +92,370 @@ func DefaultKeyMap() KeyMap {
...
+// OverwriteStyles sets the table styles, overwriting all existing styles. Note:
+// using [Model.SetStyleFunc] will override styles set in this function.
+func (m *Model) OverwriteStyles(s Styles) *Model {
+	m.styles = s
+	return m
+}
+
+// OverwriteStylesFromLipgloss sets the [Model]'s style attributes from an
+// existing [lipgloss.Table].
+func (m *Model) OverwriteStylesFromLipgloss(t *table.Table) {
```
andreynering, line 388: Not sure if we could improve the name of this function...

table/table.go
```diff
@@ -101,108 +92,370 @@ func DefaultKeyMap() KeyMap {
...
+	for _, line := range strings.Split(value, "\n") {
+		row := []string{}
+		for _, field := range strings.Split(line, separator) {
+			row = append(row, field)
+		}
+		rows = append(rows, row)
+	}
+
+	m.SetRows(rows...)
+}
```
andreynering, line 447: 1. Do we need to add this?
2. If so, would it make sense to do a proper parse as CSV, to ensure char escaping work, etc?

table/table.go
bashbunni: We could probably make it `StyleFromLipgloss` ?

table/table.go
bashbunni: I agree, I think it makes more sense to only store the data once instead of managing the same set of values in two places (Model and lipgloss Table).

I vote that we add `GetData()` that returns a `Data` then the user can call `DataToMatrix`. Performance-wise it would be the same as we would probably be doing that type conversion under the hood if we were returning `[][]string`. This might lead users to want to include non-string values in the table... Is that something we want to support? I guess that feature could sit in the backlog for a bit while we think about it.

table/table.go
bashbunni: @andreynering I believe 5 is the minimum height for the table as that would be top border + header + header border + "..." + bottom border. That's by design, but we can change that if you have a preferred behaviour in mind

table/table.go
bashbunni: Removed this!

table/table.go
andreynering: Yeah, I understand the reason. 5 is the minimum to have a minimally presentable table.

That said, if the user _really_ asked for a smaller table, I feel that we should cut the table. That's we do horizontally, btw. We have a test for width 1 on lipgloss and it looks like this:

```
┌
│
├
│
│
│
│
│
└
```

Perhaps on the TUI app the space is shared with other components, and if the window is too small, it should do what's asked.

aymanbagabas (approved):

bashbunni:
We still need to get the number of rows and/or border heights + if it has headers to be able to jump around the content with Home/End keys and PageUp/PageDown keys. @andreynering

table/table.go
```diff
@@ -101,108 +91,346 @@ func DefaultKeyMap() KeyMap {
...
+// Bubble Tea Methods
+
+// Update is the Bubble Tea [tea.Model] update loop.
 func (m Model) Update(msg tea.Msg) (Model, tea.Cmd) {
 	if !m.focus {
 		return m, nil
 	}
+	table := m.table.String()
+	// TODO make this not hard coded?
+	height := lipgloss.Height(table) - 6
```
andreynering, line 433: Note to self: take a look at this

lrstanley:
What's the progress on this? Tested it out briefly, and a bit of feedback so far:
1. `New()` returning a pointer vs `Update()` not returning a border is a little cumbersome to work with.
2. some methods return `*table.Model`, like `SetBorder()`, which is a bit confusing, as nothing else really seems to have a builder pattern.
3. updating the data at a high frequency does seem to have some performance impact. e.g. if I try to truncate the row data on my side when the height/width changes size, the app almost freezes, using 500%+ cpu.
4. there is no clean way to setup non-wrapping truncation. I can update the style to be inline, which technically does it, but doesn't do any kind of ellipsis, and makes a consecutive string of characters bridging two columns, which isn't very clean.
5. text wrapping when the screen is small also causes scrolling to break. the selected element, and the viewport don't appropriately scroll when there is an entry that is overflowing. see below video.

https://cdn.liam.sh/share/2025/07/explorer_80HMdm4hK2.mp4

there also seems to be a panic under certain scenarios (note, trace shows my fork, where all I did was merge v2-table onto current v2-exp branch, and create a tag, nothing custom):

```
Caught panic:

runtime error: index out of range [-1]

Restoring terminal...

goroutine 1 [running]:
runtime/debug.Stack()
        /usr/local/go/src/runtime/debug/stack.go:26 +0x5e
github.com/charmbracelet/bubbletea/v2.(*Program).recoverFromPanic(0xc0001a0a00, {0xabbc60, 0xc0001813f8})
        /home/liam/go/pkg/mod/github.com/charmbracelet/bubbletea/v2@v2.0.0-beta.4/tea.go:1139 +0x165
github.com/charmbracelet/bubbletea/v2.(*Program).Run.func2()
        /home/liam/go/pkg/mod/github.com/charmbracelet/bubbletea/v2@v2.0.0-beta.4/tea.go:872 +0xe8
panic({0xabbc60?, 0xc0001813f8?})
        /usr/local/go/src/runtime/panic.go:792 +0x132
github.com/charmbracelet/lipgloss/v2/table.(*StringData).At(0xc0007100c8?, 0xaff597?, 0x0?)
        /home/liam/go/pkg/mod/github.com/lrstanley/lipgloss/v2@v2.0.0-beta.4/table/rows.go:41 +0x5d
github.com/charmbracelet/lipgloss/v2/table.(*Table).constructRow(0xc0002a8008, 0xffffffffffffffff)
        /home/liam/go/pkg/mod/github.com/lrstanley/lipgloss/v2@v2.0.0-beta.4/table/table.go:498 +0x391
github.com/charmbracelet/lipgloss/v2/table.(*Table).String(0xc0002a8008)
        /home/liam/go/pkg/mod/github.com/lrstanley/lipgloss/v2@v2.0.0-beta.4/table/table.go:352 +0x7ce
github.com/charmbracelet/bubbles/v2/table.Model.View({{{{0xc0001e8c20, 0x2, 0x2}, {{...}, {...}}, 0x0}, {{0xc0001e8c40, 0x2, 0x2}, {{...}, ...}, ...}, ...}, ...})
        /home/liam/go/pkg/mod/github.com/lrstanley/bubbles/v2@v2.0.0-beta.3/table/table.go:472 +0xdf
[...]
```

PR #773, opened 2025-04-05
feat(key): add `Equal` method to compare key bindings

Kapparina:
- Implement `Equal` method to check for equality of keys and help text.
- Import `slices` package to simplify key comparison logic.

### DIFF
--- a/key/key.go
+++ b/key/key.go
@@ -36,7 +36,10 @@
 // to render help text for keystrokes in your views.
 package key
 
-import "fmt"
+import (
+	"fmt"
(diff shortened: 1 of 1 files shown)


Kapparina:
> Thank you so much for your contribution here @Kapparina!
> 
> We accidentally closed your PR after merging v2-exp. If you still think this is relevant, please re-open your changes and, we will take a look as soon as possible.

Thank you and apologies for the late reply @aymanbagabas, I have reopened this via #916.

PR #776, opened 2025-04-13
(v2) proposal(filepicker): allow configuring FS

nerg4l:
The current functionality of the file picker is limited. It would be great to have

- Help
- Filter
- Only display directories and AllowedTypes only (this could be added as a filter)
- Browsing any filesystem (even virtual)

This PR includes a proposal for the last list item. The idea is to expose an FS property which by default uses the default os filesystem but it can be changed to a different value. It seemed that this won't require a method, but I might be wrong about this.

It is a really rough attempt so please let me know how I could improve on this.

### DIFF
--- a/filepicker/filepicker.go
+++ b/filepicker/filepicker.go
@@ -4,6 +4,7 @@ package filepicker
 
 import (
 	"fmt"
+	"io/fs"
 	"os"
 	"path/filepath"
 	"sort"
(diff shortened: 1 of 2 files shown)


nerg4l:
@dolmen I think my changes are backwards compatible and can be cherry picked. However, I did not add symlink resolution, which could be a problem.

PR #797, opened 2025-06-03
(v2) Add focus status to textarea's setPromptFunc

meowgorithm:
This gives focus status to the prompt. We may want to reexamine this and add more metadata.

```go
// Before
func (m *Model) SetPromptFunc(promptWidth int, fn func(lineIndex int) string)

// After
func (m *Model) SetPromptFunc(promptWidth int, fn func(lineIndex int, focused bool) string)
```

### DIFF
--- a/textarea/textarea.go
+++ b/textarea/textarea.go
@@ -287,7 +287,7 @@ type Model struct {
 
 	// If promptFunc is set, it replaces Prompt as a generator for
 	// prompt strings at the beginning of each line.
-	promptFunc func(line int) string
+	promptFunc func(line int, focused bool) string
 
 	// promptWidth is the width of the prompt.
 	promptWidth int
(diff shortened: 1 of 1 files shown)


kujtimiihoxha (approved):

raphamorim (approved):

textarea/textarea.go
```diff
@@ -287,7 +287,7 @@ type Model struct {
 
 	// If promptFunc is set, it replaces Prompt as a generator for
 	// prompt strings at the beginning of each line.
-	promptFunc func(line int) string
+	promptFunc func(line int, focused bool) string
```
aymanbagabas, line 290: Perhaps we can have a `PromptMetadata` type that encapsulates both `lineIndex` and `focused` and can be used for any additional future metadata.

meowgorithm: Added a `PromptInfo` type to match `LineInfo`. @aymanbagabas if this looks good to you, go ahead and merge it.

aymanbagabas (approved):

PR #893, opened 2026-02-27
feat: tree bubble

### LINKED ISSUE #233: Feature request: Tree model

dlvhdr:
* Previous work: #639 
* Closes #233
* Bubble Tea example: https://github.com/charmbracelet/bubbletea/pull/1600

- [x] I have created a discussion that was approved by a maintainer (for new features).


dlvhdr:
@andreynering I see 2.1 was released. Just checking, is this planned for 2.2?

andreynering (approved):

dlvhdr:
Anything else needed here @andreynering ?

go.mod
```diff
@@ -31,3 +31,5 @@ require (
 	golang.org/x/sync v0.20.0 // indirect
 	golang.org/x/sys v0.45.0 // indirect
 )
+
+replace charm.lang/lipgloss/v2 v2.0.4 => ../lipgloss
```
andrinoff: Replace still here

tree/styles.go
```diff
@@ -0,0 +1,108 @@
...
+// component.
+func DefaultStyles(isDark bool) (s Styles) {
+	lightDark := lipgloss.LightDark(isDark)
+	verySubduedColor := lightDark(lipgloss.Color("#DDDADA"), lipgloss.Color("#3C3C3C"))
+	subduedColor := lightDark(lipgloss.Color("#9B9B9B"), lipgloss.Color("#5C5C5C"))
+
+	s.TreeStyle = lipgloss.NewStyle()
+	s.HelpStyle = lipgloss.NewStyle().PaddingTop(1)
+
+	s.NodeStyle = lipgloss.NewStyle().Foreground(lightDark(lipgloss.Color("#9B9B9B"), lipgloss.Color("b0b0b0")))
```
andrinoff: `lipgloss.Color("b0b0b0")` missing "#"

tree/tree.go
```diff
@@ -0,0 +1,741 @@
...
+
+		case key.Matches(msg, m.KeyMap.Toggle):
+			m.ToggleCurrentNode()
+		case key.Matches(msg, m.KeyMap.Open):
+			m.OpenCurrentNode()
+		case key.Matches(msg, m.KeyMap.Close):
+			m.CloseCurrentNode()
+
+		case key.Matches(msg, m.KeyMap.Quit):
+			return m, tea.Quit
```
andrinoff: why is quit of the whole TUI handled in a reusable element?

tree/node.go
```diff
@@ -0,0 +1,393 @@
...
+	t.tree.Offset(t.tree.Children().Length(), 0)
+	return t
+}
+
+// Open opens the node.
+func (t *Node) Open() *Node {
+	t.open = true
+	t.tree.Offset(0, 0)
+	return t
+}
```
andrinoff, line 89: does not clear `initialClosed`

tree/tree.go
```diff
@@ -0,0 +1,741 @@
...
+		rs := m.rootStyle()
+		m.root.RootStyle(rs)
+
+		root := ""
+		switch val := m.root.value.(type) {
+		case fmt.Stringer:
+			root = rs.Render(val.String())
+		case string:
+			root = rs.Render(ansi.Strip(val))
+		}
```
andrinoff, line 691: no `default` value. Not sure if actually needed, but your changes in `lipgloss` handle any arbitrary values

tree/tree.go
```diff
@@ -0,0 +1,741 @@
...
+		closedCharacter: m.closedCharacter,
+		treeYOffset:     m.yOffset,
+		styles:          m.styles,
+	}
+}
+
+func (m *Model) rootStyle() lipgloss.Style {
+	nodes := Nodes{m.root}
+	if m.root.yOffset == m.yOffset && m.styles.selectedNodeFunc != nil {
+		return m.styles.rootNodeFunc(nodes, 0).Inherit(
```
andrinoff, line 718: no nil guard check for `rootNodeFunc`

tree/tree.go
```diff
@@ -0,0 +1,741 @@
...
+
+// SetCursorCharacter sets the character used to represent the cursor.
+func (m *Model) SetCursorCharacter(character string) {
+	m.cursorCharacter = character
+}
+
+// SetNodes sets the tree to the given root node.
+func (m *Model) SetNodes(t *Node) {
+	m.root = t
+	m.root.value = t.Value()
```
andrinoff: this will convert a stringer in a string permanently

tree/tree.go
```diff
@@ -0,0 +1,741 @@
...
+		KeyMap:          DefaultKeyMap(),
+		openCharacter:   "▼",
+		closedCharacter: "▶",
+		cursorCharacter: "→",
+		Help:            help.New(),
+		scrollOff:       5,
+
+		showHelp: true,
+		root:     t,
+		viewport: viewport.Model{},
```
andrinoff: worth using `viewport.New()` instead, keybinds should be configurable

andrinoff: also, clearing `KeyMap`

tree/styles.go
```diff
@@ -0,0 +1,108 @@
...
+	s.nodeFunc = func(_ Nodes, _ int) lipgloss.Style {
+		return s.NodeStyle
+	}
+
+	s.SelectedNodeStyle = s.NodeStyle.Foreground(lipgloss.Color("212")).Bold(true)
```
andrinoff: no `LightDark` support

tree/styles.go
```diff
@@ -0,0 +1,108 @@
...
+		return s.SelectedNodeStyle
+	}
+
+	s.RootNodeStyle = lipgloss.NewStyle().
+		Foreground(lightDark(lipgloss.Color("#EE6FF8"), lipgloss.Color("#EE6FF8")))
+	s.rootNodeFunc = func(_ Nodes, _ int) lipgloss.Style {
+		return s.RootNodeStyle
+	}
+
+	s.ParentNodeStyle = lipgloss.NewStyle().Foreground(lipgloss.Color("99"))
```
andrinoff: no `LightDark` support

tree/styles.go
```diff
@@ -0,0 +1,108 @@
...
+	s.parentNodeFunc = func(_ Nodes, _ int) lipgloss.Style {
+		return s.ParentNodeStyle
+	}
+
+	s.CursorStyle = lipgloss.NewStyle().PaddingRight(1).Foreground(lipgloss.Color("212")).Bold(true)
```
andrinoff: no `LightDark` support

tree/styles.go
```diff
@@ -0,0 +1,108 @@
...
+	subduedColor := lightDark(lipgloss.Color("#9B9B9B"), lipgloss.Color("#5C5C5C"))
+
+	s.TreeStyle = lipgloss.NewStyle()
+	s.HelpStyle = lipgloss.NewStyle().PaddingTop(1)
+
+	s.NodeStyle = lipgloss.NewStyle().Foreground(lightDark(lipgloss.Color("#9B9B9B"), lipgloss.Color("b0b0b0")))
+	s.nodeFunc = func(_ Nodes, _ int) lipgloss.Style {
+		return s.NodeStyle
+	}
+
```
andrinoff, line 61: dead code due to `SetStyles` overriding either way? (unsure)

tree/node.go
```diff
@@ -0,0 +1,393 @@
...
+
+// Depth returns the depth of the node in the tree.
+func (t *Node) Depth() int {
+	return t.depth
+}
+
+// Size returns the number of nodes in the tree.
+// Note that if a child isn't open, its size is 1.
+func (t *Node) Size() int {
```
andrinoff, line 54: `Size()` is `len(AllNodes())`, allocating per call. This combined with `child.Size()` per child in `tree.go` is O(n^2) per `setAttributes`

andrinoff (changes requested):
Thank you for your contribution!

On top of the comments below, there's no bubble tree added in README!

go.mod
meowgorithm: Small one here: v2 is out now so let's remove this and use the latest.

tree/tree.go
dlvhdr: Hmm I don't have a good idea. I guess I copied it from https://github.com/charmbracelet/bubbles/blob/master/list/list.go?plain=1#L862.
Will remove, no reason for that.

tree/tree.go
dlvhdr: thx! I added the default case

tree/styles.go
dlvhdr: this will be overwritten only when the user calls `SetStyles` with `styles.NodeStyleFunc`, which the default don't set.

tree/node.go
dlvhdr: how does `setAttibutes` get called in this flow?
This is basically recursive calls to `tree.Children()` to figure out how many children each node has. Since `lipgloss` implemented this where if a child is collapsed, the number of children is 1, I did the same.

I figured the user would already know how many total items he put in the tree but for scrolling purposes, and understanding the vertical size of the tree, I would expose it like lipgloss.

tree/tree.go
dlvhdr: I also tested the examples I wrote and they handle quit on their own - so it works.

list/keys.go
```diff
@@ -66,6 +67,10 @@ func DefaultKeyMap() KeyMap {
 			key.WithKeys("esc"),
 			key.WithHelp("esc", "clear filter"),
 		),
+		Select: key.NewBinding(
+			key.WithKeys("v"),
+			key.WithHelp("v", "select"),
+		),
```
andrinoff: why is there a change in `list`?

tree/tree.go
```diff
@@ -0,0 +1,735 @@
...
+}
+
+func (m Model) cursorView() string {
+	if m.cursorCharacter == "" {
+		return ""
+	}
+	cursor := strings.Split(strings.Repeat(" ", m.root.Size()), "")
+	cursor[m.yOffset] = m.cursorCharacter
+	return m.styles.CursorStyle.Render(lipgloss.JoinVertical(lipgloss.Left, cursor...))
+}
```
andrinoff, line 581: draws one cursor row per node, but the tree renders one row per line. those are the same number only if every node is exactly one line tall.

tree/tree.go
```diff
@@ -0,0 +1,735 @@
...
+
+// SetCursorCharacter sets the character used to represent the cursor.
+func (m *Model) SetCursorCharacter(character string) {
+	m.cursorCharacter = character
+}
+
+// SetNodes sets the tree to the given root node.
+func (m *Model) SetNodes(t *Node) {
+	m.root = t
+	m.root.value = t.GivenValue()
```
andrinoff: not `nil` guarded. the same thing in `AllNodes` (`:582`)

tree/node.go
andrinoff: sorry, I mixed up two things. no issue with the `Size()` API. my concern is only how often it's recomputed.

`setAttributes` runs in `New`, `SetNodes` and `toggleNode`, so on every expand/collapse, and `Size()`/`AllNodes()` is hit three more times per cursor move. 

each node asks every child for its size and answering means walking that childs whole subtree, so nodes get recounted once per ancestor above them. cost grows with depth, not node count (which i mistakenly said). wide trees will be fine, but deep ones won't

tree/tree.go
```diff
@@ -0,0 +1,735 @@
...
+	if m.enumerator != nil {
+		m.root.Enumerator(*m.enumerator)
+	}
+	if m.indenter != nil {
+		m.root.Indenter(*m.indenter)
+	}
+	m.setAttributes()
+	m.updateStyles()
+	m.updateViewport(0)
+	m.setRootStyles(m.styles)
```
andrinoff: (nit) styles are installed after the viewport content is built, so we will have 1 unstyled frame. reordering will fix this!

andrinoff (changes requested):
just a little more! 

p.s. lint fails

list/keys.go
dlvhdr: I don't know :S Probably added during mvp for testing. Thanks for catching it!

tree/tree.go
dlvhdr: Hmm I haven't tested multi line nodes. I can try and support that, but I never tested that. Is that something you're interested in?
A lot of the implementation assumes this, so it will probably not be straightforward, but it is a cool idea.

tree/node.go
dlvhdr: Oh yeah I agree.
It is unfortunate. I think we can remove the `Depth` property. I didn't use it in any of the examples. In [diffnav](https://github.com/dlvhdr/diffnav) there's a config option to start the tree expanded up to a certain level. I think `diffnav` can implement this on its own rather than taxing every `tree` user with this perf penalty.

The other property that's being calculated there is the `yOffset` of every node.
This is used for methods such as:
- `(m *Model) Node(yoffset int) *Node` - to find the node at a given offset.
- `(m *Model) NodeAtCurrentOffset() *Node` - to get the node at the current offset. I use this heavily in `diffnav`.
- Styling a node if it's selected
- Exposing it to users in the `YOffset` method

I can optimize this probably by just going over the nodes below the given yOffset, but that won't really help in all cases.

I think the TL;DR is:
- I calculate this on interactions - toggling, setting new nodes etc. This makes the rendering a lot of faster as nodes can easily know how to style themselves.
- My implementation uses `yOffset` heavily and removing it will require a lot of work. Let me know if you think this is worth the time. A good use case I can test against will be good here.

dlvhdr:
@andrinoff I've fixed the small comments. For the two big comment, let me know how you want to proceed.

- We can always test the perf on a big tree to see how important it is. For my use cases in `diffnav` it works great but I admit I haven't put it to a large data set test.
- For the multi line case

tree/tree.go
andrinoff: https://github.com/user-attachments/assets/9161adbb-0764-4b28-994c-1b4fba08903d

this is a nit, but at the same time i can imagine a user creating a multi-line node, or even having word-wrapping.

It shouldn't be that hard to fix, from what i've seen

andrinoff:
let's fix multi-lines. we can leave performance as is for now, i'll still be improving performance across elements

tree/tree.go
dlvhdr: ok, I'll give it a shot.
Do note that it will make the `yOffset` calculation for each node more taxing though, since we'll need to call `lipgloss.Height` on each node.

dlvhdr: I've had at it for a couple of hours but couldn't find an elegant solution for multiline items.
When going "up" (to the previous item, not necessarily the parent) the tree, I would need to know the height of the previous item. I can maintain a bunch of pointers and maintain them and on every interaction update them. I couldn't manage to hack this.

I also tried modifying the current slow implementation of maintaning the yOffsets.
I faced an issue with scrolling the viewport. By how much should I scroll it when a line is 50 lines long? The user would probably want to scroll it independently of the tree. We'd then have to add more keybinds if we want.

Maybe there's some intuitive solution to this that I'm too tired to think of.

dlvhdr: I did end up hacking this with the existing `yOffset` implementation.
Not sure how I feel about the up/down not going to the previous/next item, but maybe it makes sense?


https://github.com/user-attachments/assets/420e76f2-b2a7-43f5-946c-58174c50e780

andrinoff: hm, it seems like a clanky hack, where it does not go to the next/previous item... I'll try something out

andrinoff:
@dlvhdr got it working!

dlvhdr:
@andrinoff just tested it on `diffnav` and it works great!

PR #909, opened 2026-03-18
fix(textarea): add ctrl+left/right for word movement

dschmidt:
## Summary

- Add `ctrl+left` and `ctrl+right` to textarea's default `WordBackward` and `WordForward` key bindings
- `textinput` already binds these keys for word movement, but `textarea` only has `alt+left`/`alt+right` — this fixes the inconsistency

## Test plan

- [x] Existing textarea tests pass
- [x] Verify `ctrl+left`/`ctrl+right` moves cursor by word in a textarea
- [x] Verify `alt+left`/`alt+right` still works as before


dschmidt:
ping? 🙃

PR #920, opened 2026-03-29
fix(textinput): use display width for cursor position with CJK characters

amosbird:
## Summary

`Cursor()` used `m.Position()` (rune index) to compute the X offset. This is incorrect for wide characters (e.g. CJK) that occupy 2 terminal columns per rune, causing the real cursor to be placed to the left of where it should be.

Fix: replace `m.Position()` with `uniseg.StringWidth(string(m.value[m.offset:m.pos]))` to compute the actual display width of the visible text before the cursor.

## Test plan

- Added `TestCursorPositionWithCJKCharacters` — pure CJK input
- Added `TestCursorPositionWithMixedASCIIAndCJK` — mixed ASCII + CJK
- Added `TestCursorPositionCJKWithOffset` — narrow width forcing scroll, cursor stays within bounds

### DIFF
--- a/textinput/textinput.go
+++ b/textinput/textinput.go
@@ -921,8 +921,8 @@ func (m Model) Cursor() *tea.Cursor {
 	w := lipgloss.Width
 
 	promptWidth := w(m.promptView())
-	xOffset := m.Position() +
-		promptWidth
+	displayWidth := uniseg.StringWidth(string(m.value[m.offset:m.pos]))
+	xOffset := displayWidth + promptWidth
 	if m.width > 0 {
 		xOffset = min(xOffset, m.width+promptWidth)
 	}
--- a/textinput/textinput_test.go
+++ b/textinput/textinput_test.go
@@ -107,6 +107,76 @@ func ExampleValidateFunc() {
 	}
 }
 
+func TestCursorPositionWithCJKCharacters(t *testing.T) {
+	t.Parallel()
+
+	ti := New()
+	ti.SetVirtualCursor(false)
+	ti.Focus()
+	ti.Prompt = "> "
+
+	// Type CJK characters that are 2 cells wide each.
+	ti = sendString(ti, "你好")
+
+	cur := ti.Cursor()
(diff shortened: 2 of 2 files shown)


ashcastle:
I reproduced this in a real downstream application, [Crush](https://github.com/charmbracelet/crush), and confirmed that this PR addresses the same mismatch.

**Environment**

- macOS 26.5.2 (Apple Silicon)
- Crush v0.76.0
- Bubbles v2.1.0
- UTF-8, `TERM=xterm-256color`

Crush calls `SetVirtualCursor(false)` for its model-picker search input. In the initial **“Find your fave”** field:

1. Type or paste `한글中文日本語ABC`.
2. The cursor drifts left by one terminal cell for every CJK character.
3. The input has 10 runes but occupies 17 cells; `Cursor()` returns a value offset of 10 instead of 17.

The issue reproduces with already-composed Unicode text, so it is not limited to IME composition.

I also ran a deterministic probe against the exact Bubbles v2.1.0 dependency:

```text
"한글 테스트" runes=6 cells=11 cursor-value-offset=6 drift=5
"中文测试" runes=4 cells=8 cursor-value-offset=4 drift=4
"日本語テスト" runes=6 cells=12 cursor-value-offset=6 drift=6
"ABC한글XYZ" runes=8 cells=10 cursor-value-offset=8 drift=2
"한글中文日本語ABC" runes=10 cells=17 cursor-value-offset=10 drift=7
```

The current Crush `nightly` branch has updated to Bubbles v2.1.1, but that release still uses `m.Position()` and remains affected.

The proposed `uniseg.StringWidth(string(m.value[m.offset:m.pos]))` calculation matches both the CJK display width and the visible horizontally-scrolled slice. I’m happy to help with rebasing or additional downstream validation if useful.

**Current-main validation:** I also cherry-picked commit `ac6f4c2` onto the current Bubbles `main` (`b52e21a`) without conflicts. `go test ./...` passes both on the original PR branch and with the change applied to current `main`.

PR #959, opened 2026-04-22
textarea: guard wordLeft against empty buffer so alt+left doesn't hang

SAY-5:
## Summary

Fixes charmbracelet/bubbletea#1652.

`wordLeft()` ran an unconditional `for {}` loop whose only exit was finding a non-space rune under the cursor. On an empty textarea (`m.value == [][]rune{{}}`, cursor at `row=0/col=0`), `characterLeft()` is a no-op and the break condition `col < len(value[row]) && !unicode.IsSpace(...)` evaluates to `0 < 0`, so the loop spins forever and freezes the entire Bubble Tea event loop the first time the user hits `alt+b` or `alt+left` on an empty input.

`doWordRight()` already has an analogous end-of-text guard. Mirror it: bail at `(row 0, col 0)` before stepping left so there is no loop with no exit.

## Tests

`TestWordLeftEmptyDoesNotHang`, runs `Update` with `alt+left` on a freshly-created textarea inside a goroutine and fails if it doesn't return within 2s. On master it hangs; with this fix it returns instantly.

```
--- PASS: TestWordLeftEmptyDoesNotHang (0.00s)
```

`go test ./textarea/...`, `go vet`, `go build` all clean.


SAY-5:
Bumping this. Repro is alt+left on an empty textarea, wordLeft loops forever because the loop condition never advances at offset 0. One line guard plus a test.

PR #1005, opened 2026-06-20
fix(textinput): account for horizontal scroll offset in Cursor()

### LINKED ISSUE #1001: v2: textinput Cursor() X ignores scroll offset when input overflows width
## Description

`textinput.Model.Cursor()` returns an X coordinate based on `m.Position()` (the absolute cursor position in the full value), but when the input text is wider than the configured width, the textinput scrolls horizontally. `View()` correctly uses `m.pos - m.offset` for the visible cursor column, but `Cursor()` does not account for the scroll offset.

This means the hardware cursor (used when `SetVirtualCursor(false)`) is placed at the wrong column when the input is scrolled.

## Version

`charm.land/bubbles/v2 v2.1.0`

## Affected code

In `textinput.go`, the `Cursor()` method:

```go
func (m Model) Cursor() *tea.Cursor {
    if m.useVirtualCursor || !m.Focused() {
        return nil
    }

    w := lipgloss.Width
    promptWidth := w(m.promptView())
    xOffset := m.Position() + promptWidth   // <-- uses absolute position
    if m.width > 0 {
        xOffset = min(xOffset, m.width+promptWidth)
    }
    // ...
}
```

Compare with `View()` which correctly adjusts for scroll:

```go
value := m.value[m.offset:m.offsetRight]
pos := max(0, m.pos-m.offset)              // <-- uses scroll-adjusted position
```

## Reproduction test

```go
package textinput_test

import (
	"strings"
	"testing"

	"charm.land/bubbles/v2/textinput"
)

func TestCursorXAccountsForScrollOffset(t *testing.T) {
	m := textinput.New()
	m.Focus()
	m.SetVirtualCursor(false)
	m.CharLimit = 200
	m.SetWidth(20)

	text := strings.Repeat("a", 30)
	m.SetValue(text)
	m.CursorEnd()

	m, _ = m.Update(nil)

	cur := m.Cursor()
	if cur == nil {
		t.Fatal("Cursor() returned nil")
	}

	width := m.Width()
	t.Logf("width=%d  pos=%d  cursor.X=%d", width, m.Position(), cur.X)

	if cur.X > width {
		t.Errorf("Cursor().X = %d, exceeds viewport width %d; "+
			"Cursor() uses Position() (absolute) instead of the "+
			"scroll-adjusted visible column", cur.X, width)
	}
}
```

**Output:**
```
width=20  pos=30  cursor.X=22
Cursor().X = 22, exceeds viewport width 20
```

## Expected behavior

`Cursor().X` should reflect the visible cursor column within the viewport (i.e., `m.pos - m.offset + promptWidth`), consistent with how `View()` renders the cursor.

## Impact

Any consumer using the hardware cursor (`SetVirtualCursor(false)`) with text longer than the input width will see the terminal cursor placed at the wrong column. We hit this building an SSH-accessible CLI where the hardware cursor is needed to avoid re-render-induced text selection breakage.

## Suggested fix

Replace `m.Position()` with `m.Position() - m.offset` in `Cursor()` (the `offset` field is unexported, so this must be fixed in the package):

```go
xOffset := (m.pos - m.offset) + promptWidth
```

guptarohit:
Fixes #1001.

`Cursor()` computed its X from the absolute `Position()`, but when the value
overflows the configured width the input scrolls horizontally. `View()` already
renders the cursor at the scroll-adjusted column (`m.pos - m.offset`); `Cursor()`
did not, so the hardware cursor (`SetVirtualCursor(false)`) landed at the wrong
column once scrolled.

This aligns `Cursor()` with `View()` by using `max(0, m.pos-m.offset) + promptWidth`.

Added `TestCursorXAccountsForScrollOffset`. Note: I used a column-accurate
assertion with the cursor scrolled into the *middle* of the viewport — the
original issue's snippet asserted `X > width`, which ignores the prompt width and
would fail even after a correct fix at end-of-input.

### DIFF
--- a/textinput/textinput.go
+++ b/textinput/textinput.go
@@ -921,8 +921,9 @@ func (m Model) Cursor() *tea.Cursor {
 	w := lipgloss.Width
 
 	promptWidth := w(m.promptView())
-	xOffset := m.Position() +
-		promptWidth
+	// Like View, use the scroll-adjusted column (pos-offset) so the cursor
+	// tracks the visible position when the value overflows the width.
+	xOffset := max(0, m.pos-m.offset) + promptWidth
 	if m.width > 0 {
 		xOffset = min(xOffset, m.width+promptWidth)
 	}
--- a/textinput/textinput_test.go
+++ b/textinput/textinput_test.go
@@ -7,6 +7,7 @@ import (
 	"testing"
 
 	tea "charm.land/bubbletea/v2"
+	"charm.land/lipgloss/v2"
 )
 
 func Test_CurrentSuggestion(t *testing.T) {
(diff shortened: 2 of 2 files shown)


thomas-mangin:
Thanks for looking into my report. Moving the cursor into the middle of the viewport is what exposes the problem, as at the end, the width clamp hides it.

The patch currently counts runes but the test expects to count cells, so it only passes for ASCII (m.pos/m.offset are rune indices)

- current patch (runes): xOffset := max(0, m.pos-m.offset) + promptWidth
- test's expected value (cells): lipgloss.Width(m.promptView()) + lipgloss.Width(string(m.value[m.offset:m.pos]))

Measuring the visible prefix in cells fixes it (w is already lipgloss.Width in Cursor()):

xOffset := w(string(m.value[m.offset:m.pos])) + promptWidth

A CJK row locks it in:
m.SetValue(strings.Repeat("世", 30)) // 2 cells/rune, so rune count != column

I used Claude to verify against v2.1.0 (width 20, prompt > , cursor mid-scroll): the rune formula gives 7 where View() renders column 12; the cell formula gives 12. So the current fix fails the test on that input.

Issue #1027, opened 2026-08-04
make input area selection only select the text


plutonium-239:
### Description

when dragging in the input area, it would be great if only the actual input got selected, instead of falling back to the terminal's selection which includes the ui as well.

andrinoff:
Hey! Thank you for your issue @plutonium-239!

Crush uses [bubbles](https://github.com/charmbracelet/bubbles)' `textarea`. transferring issue there.

Just to make sure, can you, please, describe the flow that you want with it?

plutonium-239:
I want the flow to be similar to how selecting in a regular "rendered" text works in crush (for example the LLM's output). I am able to select only the output text:
- the selection area is confined to the message area
- the selection doesn't include the UI elements like the left "current message" bar/indicator, etc.
- it has a different color than the terminal's native selection
- on copying, crush gives a "Selected text copied to clipboard" message in the status bar with a level of "OKAY!"

PR #1028, opened 2026-08-04
feat(textarea): add pointer selection

joestump:
The textarea has no selection concept and handles no mouse input, so an app embedding it cannot offer "drag to select just the input text" — a drag falls through to the terminal's own selection, which also grabs whatever UI is drawn around the textarea.

Add a pointer selection model:

  - BeginSelection / ExtendSelection / EndSelection drive a drag.
  - SelectAll, ClearSelection, HasSelection, Selection, SelectedText round out the API.
  - PositionAt maps textarea-relative coordinates to a buffer position.
  - StyleState gains a Selection style so the highlight is visually distinct from the terminal's native selection.

Coordinates are textarea-relative rather than absolute, and the drag is driven by explicit calls rather than by consuming tea.MouseMsg in Update. A component does not know where its parent drew it, so it cannot resolve absolute mouse coordinates on its own; making the caller pass local coordinates keeps that translation where the layout is known. This also matches how an app already routes clicks to a positioned child.

Mapping accounts for the prompt and line-number gutter and for soft wrapping, so clicking the second visual row of a wrapped line selects from that row rather than from the middle of the first. Out-of-range coordinates clamp to the nearest position instead of erroring.

A selection takes precedence over the inline cursor when rendering the segment that contains it, so the same cell is not styled twice. Any keyboard input clears the selection, since the buffer or cursor is about to move out from under the stored offsets.

Selecting does not modify the buffer: this covers select-and-copy, which is what an embedding app needs to read via SelectedText. Typing over a selection to replace it is deliberately left out.

Refs charmbracelet/bubbles#1027

- [ ] I have read [`CONTRIBUTING.md`](https://github.com/charmbracelet/.github/blob/main/CONTRIBUTING.md).
- [ ] I have created a discussion that was approved by a maintainer (for new features).

### DIFF
--- a/textarea/selection.go
+++ b/textarea/selection.go
@@ -1,0 +1,257 @@
+package textarea
+
+import (
+	"strconv"
+	"strings"
+
+	"github.com/rivo/uniseg"
+)
+
+// Position is a location in the textarea's logical buffer: Row indexes a
+// logical (unwrapped) line, Col indexes a rune within that line. Col may
+// equal the line length, meaning "just past the last rune".
+type Position struct {
+	Row int
+	Col int
+}
+
+// before reports whether p sorts before q in the buffer.
+func (p Position) before(q Position) bool {
+	if p.Row != q.Row {
+		return p.Row < q.Row
+	}
+	return p.Col < q.Col
+}
+
+// gutterWidth returns the number of columns rendered to the left of the text
+// content: the prompt, plus the line-number column when enabled. Mouse
+// coordinates are offset by this much before being mapped to a column.
+func (m Model) gutterWidth() int {
+	w := m.promptWidth
+	if m.ShowLineNumbers {
+		// lineNumberView renders line numbers as " %*v " where the width is
+		// the digit count of MaxHeight, so the column is digits plus the two
+		// surrounding spaces.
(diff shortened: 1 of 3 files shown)


textarea/selection.go
```diff
@@ -0,0 +1,257 @@
+package textarea
+
+import (
+	"strconv"
+	"strings"
+
+	"github.com/rivo/uniseg"
```
meowgorithm, line 7: Heads up that we've generally migrated away from `rivo/uniseg` (which is still a wonderful lib) in favor of [`clipperhouse/uax29`](https://github.com/clipperhouse/uax29) (iirc) — but I'd see if there’s something in `github.com/charmbracelet/x/ansi` you can use first because:

1. It uses our prefer libs (like `uax29`)
2. It's ansi-aware, so operations won't split ANSI sequences

/cc @andrinoff

andrinoff:
Thank you for your contribution @joestump!

Due to me also opening a PR about this (#1029) I'll take stuff that you've done better than me in this PR and move it to mine (while adding you as the author)

andrinoff:
Again, thank you for this PR, @joestump, I moved the better things to my PR, feel free to review it! Closing this

PR #1030, opened 2026-08-04
feat(textarea): add an optional line highlighter

joestump:
Feature request: #1043 — opened as an issue rather than a Discussion; happy to move it if the team prefers the Ideas board.

---

There is no way to style parts of a line in a textarea. Apps that want token highlighting in their prompt — `@file` mentions, `/command` names, anything lexical — have to either post-process the rendered string (fighting ANSI escapes the textarea just emitted) or fork the render loop wholesale.

Add a `LineHighlighter` hook:

```go
type LineHighlighter interface {
    Highlight(lineIdx int, line []rune) []lipgloss.Range
}

func (m *Model) SetHighlighter(h LineHighlighter)
```

The hook is consulted once per logical line per render and returns style ranges over that line's runes. The renderer intersects those ranges with each wrapped segment, so implementations never think about wrapping and never see a segment boundary.

Keeping the interface range-based, rather than a `func(string) string` that rewrites ANSI, means implementations never parse or preserve escape sequences, and the textarea's wrap and cursor math keeps operating on raw runes.

Neither the cell under the virtual cursor nor a selected segment is styled by ranges. The renderer already splits cursor-line content into pre- and post-cursor segments, and an active selection short-circuits the segment entirely, so no two span-styling features can fight over the same cell.

With no highlighter installed the render path is byte-identical to before — `styleSegment` short-circuits to the plain render when there are no ranges.

Two subtleties worth stating, because both are easy to get wrong:

**Grapheme clusters.** Rune offsets are converted to display cells with `ansi.StringWidth` rather than by summing per-rune widths. `lipgloss.StyleRanges` resolves the cell offsets it is handed via `ansi.Cut`, which segments by grapheme cluster. Per-rune summing disagrees on every cluster built from more than one rune — an emoji with a skin-tone modifier scores 4 instead of 2, a ZWJ family 7 instead of 2 — and the highlight then lands that many cells right of its token.

**Range order.** `StyleRanges` walks its ranges in the order given, tracking how far it has consumed. Ranges arriving out of order or overlapping make it emit part of the line twice and drop another part, so a careless highlighter would corrupt content rather than merely mis-colour it. `normalizeRanges` sorts and de-overlaps before handing them over, and the hook's contract says callers need not do it themselves.

Tests pin all three behaviours.

**One line outside the new feature**, called out so it is not missed in review: `wrappedBase` now advances by the segment length captured *before* the trailing-space trim. That space is a rune of the logical line, so counting the trimmed length shifted every later segment of that line one rune left — for the new highlight ranges and, already, for selection spans. Happy to split it into its own PR if you'd rather.

No new dependencies.

- [x] I have created a discussion that was approved by a maintainer (for new features). — filed as #1043 instead; say the word and I'll move it to Discussions.

🤖 Posted on behalf of `@joestump` by [`claude-opus-5`](https://openrouter.ai/anthropic/claude-opus-5) using [Claude Code](https://claude.com/claude-code).

### DIFF
--- a/textarea/highlighter.go
+++ b/textarea/highlighter.go
@@ -1,0 +1,115 @@
+package textarea
+
+import (
+	"charm.land/lipgloss/v2"
+	"github.com/charmbracelet/x/ansi"
+)
+
+// LineHighlighter is an optional hook that applies per-token styling to
+// rendered lines. It is consulted once per logical line per render and
+// returns style ranges over the line's runes: range Start (inclusive) and
+// End (exclusive) are rune offsets within the line, and Style is composed
+// with the line's base style by the renderer.
+//
+// The hook runs after wrapping: the textarea asks for ranges on the raw
+// logical line, then intersects those ranges with each wrapped segment at
+// render time. Implementations must therefore return ranges in rune offsets
+// of the logical line they are given — never byte offsets, and never ranges
+// derived from a different string.
+//
+// The cell under the virtual cursor is never styled by ranges: the renderer
+// splits cursor-line content into pre- and post-cursor segments and renders
+// the cursor cell itself separately.
+//
+// Keeping this interface range-based (rather than bubbline's
+// func(string) string ANSI rewriting) means implementations never have to
+// parse or preserve escape sequences, and the textarea's wrap/cursor math
+// always operates on raw runes.
+type LineHighlighter interface {
+	// Highlight returns the style ranges for the given logical line.
+	// lineIdx is the zero-based logical line number; line is the raw rune
+	// content (no trailing newline). Implementations should return nil for
+	// lines with nothing to style.
+	Highlight(lineIdx int, line []rune) []lipgloss.Range
+}
+
+// SetHighlighter installs an optional line highlighter. Pass nil to disable
+// highlighting. The highlighter is consulted at render time only; it does
+// not affect editing, wrapping, or cursor positioning.
+func (m *Model) SetHighlighter(h LineHighlighter) {
+	m.highlighter = h
+}
+
+// highlightRanges returns the highlighter's ranges for a logical line, or
+// nil when no highlighter is installed.
+func (m *Model) highlightRanges(lineIdx int, line []rune) []lipgloss.Range {
+	if m.highlighter == nil {
+		return nil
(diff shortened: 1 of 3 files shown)


joestump-agent (commented):
Verified by checking out the branch and running the full gate: `go test ./...`, `go vet ./textarea/...`, `gofmt -l` — all green, no regressions outside `textarea`. Also traced the render path in `view()` by hand: the cursor-cell split and the selection short-circuit do prevent ranges from ever competing with the cursor or selection styling, as claimed.

The two subtleties the PR calls out are real and correctly handled — I specifically checked:

- **Grapheme clusters:** `runeDisplayWidth` via `ansi.StringWidth` matches how `lipgloss.StyleRanges` resolves offsets (`ansi.Cut`, grapheme-segmented). The `TestRuneDisplayWidthMatchesGraphemeSegmentation` cases (skin-tone modifier, ZWJ family) pin the exact disagreement per-rune summing would introduce.
- **Range order:** `normalizeRanges` sorting + de-overlapping before `StyleRanges` is the right guard given its consumed-offset walk, and making it the renderer's job rather than each implementation's is the right contract.

Minor observations, none blocking:

1. **The `wrappedBase`/`segLen` fix changes existing behavior** (selection spans on lines where a wrapped segment had its trailing space trimmed). It's a genuine bug fix, but it's bundled with a new feature. I'd take the author up on the standing offer to split it into its own PR — it's exactly the kind of one-line behavior change that deserves its own bisectable commit and changelog entry.
2. `highlightRanges` is a one-line nil-check wrapper around `m.highlighter.Highlight`; inlining the nil check at the call site (or dropping the helper) would be slightly more idiomatic, but it's fine as-is.
3. The hook is invoked once per logical line per render with no caching hook. For a syntax-highlighting implementation that re-lexes on every frame this could get hot on large buffers — worth a note in the docs eventually, not a blocker.

Nice design call keeping the interface range-based rather than `func(string) string` — it keeps ANSI handling entirely inside the renderer.

🤖 This was posted autonomously by [`glm-5.3`](https://openrouter.ai/z-ai/glm-5.3) using [Crush](https://github.com/charmbracelet/crush).

PR #1031, opened 2026-08-04
feat(textarea): add ByteOffset and SetCursorByteOffset

joestump:
Feature request: #1044 — opened as an issue rather than a Discussion; happy to move it if the team prefers the Ideas board.

---

`Line()` and `Column()` locate the cursor in the value's grid of rows and runes. That is the wrong coordinate space for a caller that wants to treat `Value()` as a flat string — slicing around the cursor, scanning backwards for a token boundary, splicing a completion in. Converting grid coordinates to a byte index means re-deriving the join `Value()` already performed, and every caller writes the same loop.

Add the pair:

```go
func (m Model) ByteOffset() int
func (m *Model) SetCursorByteOffset(off int)
```

One offset is not a position the cursor can occupy: one landing inside a multi-byte rune. `SetCursorByteOffset` snaps it forward to the next one that is, so round-tripping an offset from `ByteOffset` is always exact. A negative offset clamps to the start, a past-the-end offset to the end.

The `"\n"` `Value()` inserts between rows belongs to no row, but it costs no position either: the offset of that byte is the end of the row before it, and the offset after it is the start of the row that follows. Both are ordinary cursor positions. A test covers each side.

Getting the mid-rune case wrong is quiet and surprising: slicing the row at the requested byte yields a partial UTF-8 sequence, which decodes to U+FFFD and counts as a whole rune, so asking for byte 2 of `"世界"` lands the cursor at byte 6 — past the rune the caller was pointing into, not before it. Counting rune starts up to the offset instead lands it at byte 3. A test pins this.

`SetCursorByteOffset` repositions the viewport like every other cursor mover, so restoring a saved offset after an edit cannot leave the cursor scrolled out of sight.

No new dependencies, and no change to existing behaviour — two new methods in a new file.

- [x] I have created a discussion that was approved by a maintainer (for new features). — filed as #1044 instead; say the word and I'll move it to Discussions.

🤖 Posted on behalf of `@joestump` by [`claude-opus-5`](https://openrouter.ai/anthropic/claude-opus-5) using [Claude Code](https://claude.com/claude-code).

### DIFF
--- a/textarea/byteoffset.go
+++ b/textarea/byteoffset.go
@@ -1,0 +1,65 @@
+package textarea
+
+// ByteOffset returns the cursor's position as a byte offset into [Model.Value].
+//
+// [Model.Line] and [Model.Column] locate the cursor within the value's grid of
+// rows and runes, which is the wrong coordinate space for a caller that wants
+// to slice, splice, or scan the value as a flat string. Value joins the rows
+// with "\n", so the offset is the byte length of every preceding row plus its
+// separator, then the current row up to the cursor.
+//
+// The returned offset always lands on a UTF-8 boundary.
+func (m Model) ByteOffset() int {
+	var off int
+	for i, row := range m.value {
+		if i == m.row {
+			return off + len(string(row[:clamp(m.col, 0, len(row))]))
+		}
+		off += len(string(row)) + 1 // +1 for the joining newline
+	}
+	return off
+}
+
+// SetCursorByteOffset places the cursor at a byte offset into [Model.Value],
+// the inverse of [Model.ByteOffset]. A negative offset clamps to the start and
+// an offset past the end clamps to the end.
+//
+// An offset landing inside a multi-byte rune, or on the "\n" that Value
+// inserts between rows, is not a position the cursor can occupy; it snaps
+// forward to the next one that is. Round-tripping an offset produced by
+// ByteOffset is therefore always exact.
+func (m *Model) SetCursorByteOffset(off int) {
+	if off < 0 {
+		off = 0
+	}
+
+	var seen int
+	for row, line := range m.value {
+		s := string(line)
+		if off <= seen+len(s) {
+			m.row = row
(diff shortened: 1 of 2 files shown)


joestump-agent (commented):
Verified by checking out the branch and running the full gate: `go test ./...`, `go vet ./textarea/...`, `gofmt -l` — all green. No changes outside the two new files, as claimed.

The two behavioral edges are both correct and both pinned by tests:

- **Mid-rune snap-forward:** counting rune starts up to the offset (rather than slicing at it) lands the cursor at the *start* of the rune the offset points into, which is the useful direction — byte 1 of `"世界"` → column 1 / byte 3, not column 0. `TestSetCursorByteOffsetMidRuneSnapsForward` covers both interior bytes.
- **Row separators:** offsets on either side of a joining `"\n"` map to distinct ordinary positions (end of row N, start of row N+1) — `TestSetCursorByteOffsetAtRowSeparator` covers both. Round-trip exactness is exhaustively checked over a value mixing CJK, an emoji with skin-tone modifier, and an empty row.

One small nit, non-blocking:

- In `SetCursorByteOffset`, the `if off-seen >= len(s) { col = len([]rune(s)) }` branch after the rune-counting loop is dead code: when `off-seen >= len(s)` the loop's break condition never fires, so `col` already equals the row's rune count. It can simply be deleted.

Also worth a maintainer's eye: `ByteOffset()` returns the accumulated `off` unchanged if `m.row` is somehow past the end of `m.value` (the loop falls through). I couldn't construct that state through the public API, but if the maintainers know of one, a defensive clamp would make the "always a valid UTF-8 boundary into `Value()`" doc promise unconditional.

Well-scoped, well-tested pair of methods — exactly the coordinate-space conversion every caller of `Value()` currently re-implements.

🤖 This was posted autonomously by [`glm-5.3`](https://openrouter.ai/z-ai/glm-5.3) using [Crush](https://github.com/charmbracelet/crush).

PR #1036, opened 2026-08-15
fix(textarea): stop word-left at input boundary

### LINKED ISSUE #1652: bug: textarea infinite loop when pressing alt+left on empty input
**Describe the bug**

Pressing `alt+b` or `alt+left` (the `WordBackward` binding) on an empty `textarea.Model` causes an infinite loop that freezes the entire Bubble Tea event loop. The UI becomes completely unresponsive.

`wordLeft()` in `textarea/textarea.go:923-929` has an unconditional `for {}` loop whose only exit is finding a non-space rune under the cursor. When the textarea is empty (`m.value` is `[][]rune{ {} }`, cursor at `row=0, col=0`), `characterLeft(true)` is a no-op and the break condition `col < len(value[row])` evaluates to `0 < 0` → false, so the loop spins forever.

Note: `doWordRight()` already handles this correctly — it checks for end-of-text before looping.

**Setup**

- `charm.land/bubbles/v2 v2.0.0`
- OS: macOS
- Shell: zsh
- Terminal emulator: Ghostty
- Terminal multiplexer: N/A

**To Reproduce**

1. Create a `textarea.Model` with default (empty) content
2. Focus the textarea
3. Without typing anything, press `alt+b` or `alt+left`
4. The UI freezes indefinitely

**Source Code**

```go
package main

import (
	"fmt"
	"time"

	"charm.land/bubbles/v2/textarea"
	tea "charm.land/bubbletea/v2"
)

type model struct {
	ta textarea.Model
}

	return m.ta.Focus()
}

	var cmd tea.Cmd
	m.ta, cmd = m.ta.Update(msg)
	return m, cmd
}

func (m model) View() tea.View {
	return tea.NewView(fmt.Sprintf(
		"Press alt+b or alt+left to freeze.\n\n%s",
		m.ta.View(),
	))
}

func main() {
	ta := textarea.New()
	ta.SetHeight(3)
	ta.SetWidth(40)
	ta.Placeholder = "type something, or don't..."

	p := tea.NewProgram(model{ta: ta})
	go func() {
		time.Sleep(10 * time.Second)
		fmt.Println("\n[timeout] program was stuck — killed")
		p.Kill()
	}()
		fmt.Println("error:", err)
	}
}
```

**Expected behavior**

Pressing `alt+b` / `alt+left` on an empty textarea should be a no-op (there is no word to the left of position 0,0), not freeze the program.

**Screenshots**

N/A

**Additional context**

Suggested fix — add a boundary check at the top of `wordLeft()`, mirroring what `doWordRight()` already does:

```go
func (m *Model) wordLeft() {
	if m.row == 0 && m.col == 0 {
		return
	}
	// ... rest unchanged
}
```

OxQuasar:
## Summary

- stop `textarea.wordLeft` when `characterLeft` cannot move the cursor
- cover empty input, leading whitespace, whitespace-only preceding lines, and normal word movement

## Problem

`wordLeft` loops until it finds a non-space rune. At the beginning of the input, `characterLeft` is a no-op, so empty input or a whitespace-only prefix makes the loop spin forever and freezes the Bubble Tea event loop.

Checking only the cursor position before entering the loop is insufficient: when navigation starts inside leading whitespace, the cursor reaches `(0, 0)` during traversal. Detecting lack of progress after each attempted movement covers both cases and directly enforces the loop's termination invariant.

Related to #948, #987, and #1019. Fixes charmbracelet/bubbletea#1652.

## Testing

The regression test was run before the fix and timed out for:

- empty input
- the cursor already at the buffer start
- the cursor inside leading whitespace
- a multiline whitespace-only prefix

After the fix, the following pass:

- `go test ./...`
- `go vet ./...`
- `go test -race -covermode=atomic -coverprofile=profile.cov ./...`
- `golangci-lint v2.9 run` (0 issues)


andrinoff (approved):
Oh, this is such a valid issue, thank you for the fix!

